PR #10, opened 2022-10-14
Construct L2 images from L1 images via ramp fitting.

schlafly:
This PR adds ramp fitting routines and constructs L2 images from L1 images using them.  It also adds the corresponding docs.


schlafly:
As part of trying to improve the docs I learned that if I use numpydoc & stsci-rtd-theme, the separation between the parameter names and types disappears making the docs hard to read.  I've "fixed" this for the moment by moving to sphinx-rtd-theme, but that's obviously not ideal.  But it will work for now.

PR #13, opened 2022-10-26
move build configuration into `pyproject.toml`

zacharyburnett:
`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Support for the `tool.setuptools` table will likely be in beta until `toml` is natively supported with the release of Python 3.11.

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


schlafly:
Thanks.  I have no experience with this kind of packaging and will defer to you.  Should we go ahead and merge this?

zacharyburnett:
> Thanks. I have no experience with this kind of packaging and will defer to you. Should we go ahead and merge this?

I'd wait until the unit tests are in so we can test it, but sure

schlafly (approved):

PR #14, opened 2022-10-26
Rcal-298: Set Up Testing Infrastructure

PaulHuwe:
This ticket sets up the GitHub actions for RomanISim. The actions are a straight port of the ones form RomanCAL. In addition, I modified the code to pass all the checks (sans test coverage - that will be delineated in other tickets).


PaulHuwe:
I left significant code changes I made in comments, which I can clean up before transition from draft PR to full PR. Please let me know if you guys see anything that should be changed, tests that should be excised, etc.

romanisim/util.py
```diff
@@ -194,7 +194,7 @@ def unflatten_value(k, v):
             if 'file' in k:
                 v = stnode.FileDate(v)
         except Exception as e:
-            pass
+            raise Exception(e)
```
schlafly, line 197: I think this change will actually have an effect.  I forget all of the different kinds of objects that could be in one of these metadata files, and which types of keys caused problems.  But I got tired of tracking them down and they weren't needed for the CRDS queries, so I ended up just catching all exceptions here and ignoring them.  The approach you once suggested re doing this through a image model metadata structure is likely a better approach, but let's not change this without actually fixing this here.

schlafly:
This all looks good to me.  I added a comment about one change I worry will break things, where I intentionally suppresed some exceptions when converting some metadata between the kinds than CRDS accepts and the kinds in the image model metadata.  I'm sure there's a better way to handle the metadata passing, but I don't want to break the current hack-y functionality with this PR.  You can go ahead and delete all the commented code as far as I can tell.

schlafly (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,237 @@
...
+      - '*.x'
+    tags:
+      - '*'
+  pull_request:
+    branches:
+      - main
+  schedule:
+    # Weekly Monday 9AM build
+    # * is a special character in YAML so you have to quote this string
+    - cron: '0 9 * * 1'
```
ddavis-stsci, line 16: If you are going to run this once a week I'd suggest maybe Sunday at 9am. There can be a lot of activity of the regression machines on Monday.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,237 @@
...
+    needs: [ test ]
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest ]
+        python: [ 3.8, 3.9, '3.10' ]
+        numpy: [ '1.18.*', '1.19.*', '1.20.*', '1.21.*', '1.22.*' ]
+        exclude:
```
ddavis-stsci, line 176: I'd consider dropping support numpy 3.18 and maybe 3.19?

ddavis-stsci (approved):
Overall LGTM

.github/workflows/ci.yml
WilliamJamieson: This is a github workflow so it should be fine.

.github/workflows/ci.yml
WilliamJamieson: Astropy is dropping support in the next release for numpy 1.18 and 1.19, so we should follow the same path here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,237 @@
...
+  CRDS_PATH: "/tmp/crds_cache"
+  CRDS_CLIENT_RETRY_COUNT: 3
+  CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
+
+jobs:
+  style:
+    name: Code style checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.0.2
```
WilliamJamieson, line 29:
```suggestion
      - uses: actions/checkout@v3
```

I would only pin to the major version (no minor, or bugfix version pins). @zacharyburnett what do you think?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,237 @@
...
+        id: python
+        with:
+          python-version: 3.9
+      - uses: actions/cache@v3.0.5
+        with:
+          path: ${{ env.pythonLocation }}
+          key: style-${{ runner.os }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/setup.*') }}
+      - run: pip install pyproject-flake8
+      - run: pip freeze
+      - run: pflake8
```
WilliamJamieson, line 40: Why not use `tox` environments?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,237 @@
...
+          python-version: 3.9
+      - uses: actions/cache@v3.0.5
+        with:
+          path: ${{ env.pythonLocation }}
+          key: test-coverage-${{ runner.os }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/setup.*') }}
+      - run: pip install ".[test]" pytest-xdist pytest-cov
+      - run: pip freeze
+      - run: pytest -n auto --pyargs romanisim --cov-report xml --cov
+      - run: coverage report -m
+      - uses: codecov/codecov-action@master
```
WilliamJamieson, line 215:
```suggestion
      - uses: codecov/codecov-action@v3
```

This action uses a main branch not a master branch, which is why the CI is failing right now. It is better to pin a major release version, the latest appears to be `v3`.

romanisim/apt.py
```diff
@@ -6,8 +6,9 @@
 fraction of what an APT file seems able to do.
 """
 
-import xml
-from xml.etree import ElementTree
+# import xml
+# from xml.etree import ElementTree
```
WilliamJamieson, line 10:
```suggestion
```

No need to leave commented out code floating around. You can find the code in the git history.

romanisim/apt.py
```diff
@@ -48,7 +49,8 @@ def read_apt(filename):
     """
     # I don't know anything about reading XML.
     # In general it's very flexible and can do anything.
-    tree = ElementTree.parse(filename)
+    # tree = ElementTree.parse(filename)
```
WilliamJamieson, line 52:
```suggestion
```

Another comment

romanisim/image.py
```diff
@@ -5,7 +5,7 @@
 """
 
 import time
-import logging
+# import logging
```
WilliamJamieson, line 8:
```suggestion
```

romanisim/image.py
```diff
@@ -17,15 +17,15 @@
 import roman_datamodels.testing.utils
 from . import wcs
 from . import catalog
-from . import bandpass
-from . import psf
+# from . import bandpass
+# from . import psf
```
WilliamJamieson, line 21:
```suggestion
```

romanisim/image.py
```diff
@@ -17,15 +17,15 @@
...
 from . import parameters
 from . import util
 import romanisim.l1
 import romanisim.bandpass
 import romanisim.psf
 from romanisim import log
-from romanisim import parameters
+# from romanisim import parameters
```
WilliamJamieson, line 28:
```suggestion
```

romanisim/image.py
```diff
@@ -393,7 +393,7 @@ def simulate(metadata, objlist,
         gain = galsim.roman.gain
         flat = 1
 
-    out = dict()
+    # out = dict()
```
WilliamJamieson, line 396:
```suggestion
```

romanisim/l1.py
```diff
@@ -104,7 +109,7 @@
 import numpy as np
 import asdf
 import galsim
-from galsim import roman
+# from galsim import roman
```
WilliamJamieson, line 112:
```suggestion
```

romanisim/ramp.py
```diff
@@ -104,7 +104,7 @@ def construct_covar(read_noise, flux, ma_table):
         covariance matrix of first finite differences of unevenly sampled
         resultants.
     """
-    read_time = parameters.read_time
+    # read_time = parameters.read_time
```
WilliamJamieson, line 107:
```suggestion
```

romanisim/util.py
zacharyburnett: I agree, if you are raising a general exception then you should get rid of the `try` block

romanisim/ramp.py
```diff
@@ -315,8 +315,8 @@ def ki(self, flux, read_noise):
             ramp pixels
         """
         # clip points outside range to edges.
-        mingrid = self.flux_on_readvar_pts[0]
-        maxgrid = self.flux_on_readvar_pts[-1]
+        # mingrid = self.flux_on_readvar_pts[0]
+        # maxgrid = self.flux_on_readvar_pts[-1]
```
WilliamJamieson, line 319:
```suggestion
```

romanisim/ramp.py
```diff
@@ -341,16 +341,16 @@ def variances(self, flux, read_noise):
             and the total noise
         """
         # clip points outside range to edges.
-        mingrid = self.flux_on_readvar_pts[0]
-        maxgrid = self.flux_on_readvar_pts[-1]
+        # mingrid = self.flux_on_readvar_pts[0]
+        # maxgrid = self.flux_on_readvar_pts[-1]
```
WilliamJamieson, line 345:
```suggestion
```

romanisim/ramp.py
```diff
@@ -444,15 +444,13 @@ def simulate_many_ramps(ntrial=100, flux=100, readnoise=5, ma_table=None):
         simulated resultants
     """
     from . import l1
-    read_time = parameters.read_time
+    # read_time = parameters.read_time
```
WilliamJamieson, line 447:
```suggestion
```

romanisim/wcs.py
```diff
@@ -124,7 +125,7 @@ def make_wcs(targ_pos, roll_ref, distortion, wrap_v2_at=180, wrap_lon_at=360):
     # roll_ref for each SCA, and how that would best be done; if nothing else,
     # we do some finite differences to get the direction +V3 on the sky and
     # compute an angle wrt north.
-    v2_ref = v3_ref = 0
+    # v2_ref = v3_ref = 0
```
WilliamJamieson, line 128:
```suggestion
```

romanisim/wcs.py
```diff
@@ -191,9 +192,19 @@ def _radec(self, x, y, color=None):
         if np.ndim(x) == np.ndim(y) == 0:
             return r[0], d[0]
         else:
-            assert np.ndim(x) == np.ndim(y)
-            assert x.shape == y.shape
-            return r, d
+            # assert np.ndim(x) == np.ndim(y)
+            # assert x.shape == y.shape
```
WilliamJamieson, line 196:
```suggestion
```

romanisim/wcs.py
```diff
@@ -207,9 +218,17 @@ def _xy(self, ra, dec, color=None):
         if np.ndim(ra) == np.ndim(dec) == 0:
             return x[0], y[0]
         else:
-            assert np.ndim(ra) == np.ndim(dec)
-            assert ra.shape == dec.shape
-            return x, y
+            # assert np.ndim(ra) == np.ndim(dec)
+            # assert ra.shape == dec.shape
+            # return x, y
```
WilliamJamieson, line 223:
```suggestion
```

setup.py
```diff
@@ -1,10 +1,12 @@
...
-import sys
-from setuptools import setup, find_packages, Extension, Command
-from setuptools.command.test import test as TestCommand
+
+# import os
+# import pkgutil
+# import subprocess
+# import sys
+# from setuptools import setup, find_packages, Extension, Command
+# from setuptools.command.test import test as TestCommand
```
WilliamJamieson, line 8:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,237 @@
...
+        os: [ ubuntu-latest ]
+        python: [ 3.8, 3.9, '3.10' ]
+        numpy: [ '1.18.*', '1.19.*', '1.20.*', '1.21.*', '1.22.*' ]
+        exclude:
+          - python: '3.10'
+            numpy: '1.18.*'
+          - python: '3.10'
+            numpy: '1.19.*'
+          - python: '3.10'
+            numpy: '1.20.*'
```
zacharyburnett, line 182:
```suggestion
        numpy: [ '1.20.*', '1.21.*', '1.22.*' ]
        exclude:
          - python: '3.10'
            numpy: '1.20.*'
```

setup.py
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
```
WilliamJamieson, line 1: Switching to a pure `pyproject.toml` configuration instead of using `setup.cfg` and `setup.py` should be considered.

.github/workflows/ci.yml
zacharyburnett: Yes this is my fault, sorry, I introduced a fix for this in https://github.com/spacetelescope/romancal/pull/592

romanisim/util.py
schlafly: Let's just leave this as a blanket suppression of the possible exceptions here for now.  I agree this is bad practice, but the proposed change will break things and RDM-388 proposes a better way to handle this in general.  Or we can imagine a separate PR that does the translation between CRDS metadata and roman data model metadata in both directions.  But let's not mix that with this PR.

WilliamJamieson: Why not have it just turn the errors into warnings at the very least? That way you have an idea of what is going on.

.github/workflows/ci.yml
zacharyburnett: yes indeed, it would be better to be `v3`

.github/workflows/ci.yml
zacharyburnett, line 40: this should be `flake8`, I have another PR to address this issue in `romancal`
```suggestion
      - run: pip install flake8
      - run: pip freeze
      - run: flake8
```

setup.py
zacharyburnett: https://github.com/spacetelescope/romanisim/pull/13

.github/workflows/ci.yml
zacharyburnett: https://github.com/spacetelescope/romanisim/pull/14#discussion_r1006877851

WilliamJamieson: In general, IMOP we should be providing `tox` envs for the CI to run, rather than running a bunch of commands. This is because `tox` builds an exactly reproducible python environment for testing and then runs an identical set of commands.

zacharyburnett: That is true, though is does add another level of complexity it does allow the user to reproduce the tests locally for their Python version

romanisim/wcs.py
```diff
@@ -191,9 +192,19 @@ def _radec(self, x, y, color=None):
...
+                return r, d
+            else:
+                if (np.ndim(x) != np.ndim(y)):
+                    raise Exception("np.ndim(x) != np.ndim(y) => " + str(
+                        np.ndim(x)) + " != " + str(np.ndim(y)))
+                elif (x.shape != y.shape):
+                    raise Exception("x.shape != y.shape => " + str(
+                        x.shape) + " != " + str(y.shape))
+                else:
+                    raise Exception("Invalid x & y.")
```
schlafly, line 207: This is a reasonable improvement, but for what it's worth, the original code is galsim boilerplate rather than anything original.  Likewise for the related _xy stanza.

zacharyburnett, line 207: I would suggest you use a specific exception:
```suggestion
                    raise ValueError(f"np.ndim(x) != np.ndim(y) => {np.ndim(x)} != {np.ndim(y)}")
                elif (x.shape != y.shape):
                    raise ValueError(f"x.shape != y.shape => {x.shape} != {y.shape}")
                else:
                    raise ValueError("Invalid x & y.")
```

.github/workflows/ci.yml
PaulHuwe: No harm in changing it (and may be impactful in the future), so I did so.

.github/workflows/ci.yml
PaulHuwe: Removed.

.github/workflows/ci.yml
PaulHuwe: Done

.github/workflows/ci.yml
PaulHuwe: Done

.github/workflows/ci.yml
PaulHuwe: Changed.

romanisim/wcs.py
PaulHuwe: The asserts were causing bandit errors, hence this rewrite.

PaulHuwe: Changed in both blocks.

.github/workflows/ci.yml
PaulHuwe: Fixed.

.github/workflows/ci.yml
PaulHuwe: Done

.github/workflows/ci.yml
PaulHuwe: I leave this to Zach's ticket.

zacharyburnett: I'll make a separate PR to implement tox

romanisim/apt.py
PaulHuwe: Cleaned.

PaulHuwe:
@zacharyburnett @WilliamJamieson The codecoverage report is still not working in the environment, though the commands run locally. Any ideas why that would be?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,233 @@
...
+        id: python
+        with:
+          python-version: 3.9
+      - uses: actions/cache@v3
+        with:
+          path: ${{ env.pythonLocation }}
+          key: test-coverage-${{ runner.os }}-${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/setup.*') }}
+      - run: pip install ".[test]" pytest-xdist pytest-cov
+      - run: pip freeze
+      - run: pytest -n auto --pyargs romanisim --cov-report xml --cov
```
zacharyburnett, line 209:
```suggestion
      - run: pytest -n auto --cov-report xml --cov-report term-missing --cov
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,233 @@
...
+      - run: coverage report -m
```
zacharyburnett, line 210:
```suggestion
      - run: pytest -n auto --cov-report xml --cov-report term-missing --cov
```

zacharyburnett:
> @zacharyburnett @WilliamJamieson The codecoverage report is still not working in the environment, though the commands run locally. Any ideas why that would be?

I tested removing `--pyargs` and adding a `term-missing` report to the output, it seems to work

zacharyburnett (approved):

nden:
🚀

PR #16, opened 2022-11-02
Implement classical non-linearity

schlafly:
This adds a new module that handles classical non-linearity.

There's some complexity needed to handle both images that already include poisson noise (e.g., from galsim's photon shooting rendering approach) and non-linearity.  That took a while to get right but ends up not being very expensive.  There is documentation in the l1 documentation and modules.

Also fix a bug in the updated exception logic in GWCS._xy that was preventing any images from being made when using CRDS.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,6 +2,7 @@
 
 [flake8]
 select = F, W, E, C
+ignore = W503
 # We should set max line length lower eventually
 max-line-length = 130
 exclude =
--- a/docs/romanisim/l1.rst
+++ b/docs/romanisim/l1.rst
@@ -12,7 +12,7 @@ This process proceeds by simulating each read, drawing the appropriate number of
 After apportioning counts to resultants, systematic effects are added to the resultants.  Presently only read noise is added.  The read noise is averaged down like :math:`1/\sqrt{N}`, where :math:`N` is the number of reads contributing to the resultant.
 
 Nonlinearity
-============
+------------
 
 Non-linearity is considered when L1 images are constructed and a non-linearity model is provided (e.g., from CRDS).  Non-linearity is implemented by including an additional "efficiency" parameter that is the linearity-affected count rate divided by the idealized count rate absent nonlinearity; this is the reciprocal of the derivative of the non-linearity correction function.  Two complications arise:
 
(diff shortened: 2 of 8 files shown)


schlafly:
@PaulHuwe , @zacharyburnett , I added W503 to the flake8 ignore section because I couldn't figure out how to deal with not being allowed to break a line before (W503) or after (W504) a binary operator.  If this is too repulsive please let me know what is appropriate.

zacharyburnett:
I think these errors can be solved by adding `build` to the `norecursedirs` list in `pyproject.toml`

schlafly:
@PaulHuwe , I'm failing lots of tests but I don't really understand; our tests don't cover anything and I didn't consciously make any changes that touch any tests?

zacharyburnett:
@schlafly yes it appears that `pytest` discovery is getting confused by the `build` directory, so you should add it to the `norecursedirs`

schlafly:
@zacharyburnett , it looks like this builds now, and I'll go ahead and merge.  Thanks.  The code coverage reports that I've increased coverage by 19%, despite not touching any tests or doing anything.  That feels broken?

zacharyburnett:
> @zacharyburnett , it looks like this builds now, and I'll go ahead and merge. Thanks. The code coverage reports that I've increased coverage by 19%, despite not touching any tests or doing anything. That feels broken?

yes that is strange, the coverage was 1% and now is 20%. However I can't see any actual issues with the tests themselves, and the coverage modules seems to be working fine from what I can see in the report

PaulHuwe:
> @PaulHuwe , @zacharyburnett , I added W503 to the flake8 ignore section because I couldn't figure out how to deal with not being allowed to break a line before (W503) or after (W504) a binary operator. If this is too repulsive please let me know what is appropriate.

- I fixed lots of these in my tickets. I am not religious about this particular warning.

PaulHuwe:
> > @zacharyburnett , it looks like this builds now, and I'll go ahead and merge. Thanks. The code coverage reports that I've increased coverage by 19%, despite not touching any tests or doing anything. That feels broken?
> 
> yes that is strange, the coverage was 1% and now is 20%. However I can't see any actual issues with the tests themselves, and the coverage modules seems to be working fine from what I can see in the report

This appears to be oddly common? I have seen it several times over several projects, but never investigated it further.

PaulHuwe (approved):
LGTM

PR #18, opened 2022-11-08
Subtract dark before ramp fitting.

schlafly:
We were successfully including dark current in the L1 images, but had forgotten to subtract it before ramp fitting when constructing L2 images.  This PR corrects that oversight.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -55,7 +55,7 @@ import crds
 
 
 def make_l2(resultants, ma_table, read_noise=None, gain=None, flat=None,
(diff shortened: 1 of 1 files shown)


schlafly:
@PaulHuwe , any interest in reviewing this?  I think it's pretty straightforward, though tests are "failing" because image.py doesn't currently have unit tests---though I'm getting there.  I would have just merged this straight away, but maybe we're getting far enough along that we should try to formalize this a bit.

PaulHuwe (approved):
LGTM

PR #19, opened 2022-11-08
Some simple unit tests on the util module.

schlafly:
This adds some simple unit tests on the util module.


schlafly:
@PaulHuwe , @nden, can you let me know if this is what you're expecting unit tests on these modules to look like?

PaulHuwe (commented):
Outside of my one comment, it looks fine.

PR #22, opened 2022-11-22
Add ramp fitting tests.

schlafly:
Add tests of ramp fitting routines.
Fix minor issues in l1 and ramp routines that don't broadcast well that came up when adding ramp fitting tests.

### DIFF
--- a/romanisim/l1.py
+++ b/romanisim/l1.py
@@ -167,12 +167,12 @@ def tij_to_pij(tij, remaining=False):
     for resultant in tij:
         pi = []
         for t in resultant:
-            pi.append((t - tlast) / tremaining)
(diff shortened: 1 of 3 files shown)


schlafly:
Okay, this ramp fitting set of unit tests looks pretty good to me now.  I had to loosen the tolerances for the MacOS unit tests relative to the ubuntu unit tests, but the tolerances are still quite tight.  One test is stochastic and relies on the chi^2 of a distribution being close to expectation.  It's possible that some day fails without any code change, but I set the limits lax enough that that will happen very rarely.  i.e., goal is not in our lifetimes.

PaulHuwe (approved):
LGTM

PR #24, opened 2022-11-30
RCAL-430: Badpass Unit Tests

PaulHuwe:
Added unit tests for bandpass. Added a comment to bandpass for clarification.

### DIFF
--- a/romanisim/bandpass.py
+++ b/romanisim/bandpass.py
@@ -114,6 +114,9 @@ def get_abflux(bandpass):
         the zero point flux (photons / s)s
     """
     bandpass = galsim2roman_bandpass.get(bandpass, bandpass)
+
+    # If abflux for this bandpass has been calculated, return the calculated
+    # value instead of rerunning an expensive calculation
     abflux = getattr(get_abflux, 'abflux', None)
     if abflux is None:
         abflux = compute_abflux()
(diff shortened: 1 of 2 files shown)


schlafly:
I'm trying to think of what the most useful unit tests would be.  It would be nice to come closer to having a test that verifies correctness.

In the limit that the filter curve is a delta function, one could analytically compute the zero point, which could then be compared with the computed zero point?  That would be a good test, though we wouldn't expect perfect agreement between the numerical and analytic integrals---but we should be able to do <~1 mmag.

At present I expect that every time we update the effective area file we will also have to update the tests.  I imagine that the final zero points are still uncertain at the 10% level (making something up!), so I don't think I would make the tests on particular values more stringent than that?

PaulHuwe:
I will look at implementing both of these. I do think it is good to test the default effective area file if it will see use, so I will loosen the bounds on that fit (and maybe add more).

> In the limit that the filter curve is a delta function, one could analytically compute the zero point, which could then be compared with the computed zero point? That would be a good test, though we wouldn't expect perfect agreement between the numerical and analytic integrals---but we should be able to do <~1 mmag.
> 
> At present I expect that every time we update the effective area file we will also have to update the tests. I imagine that the final zero points are still uncertain at the 10% level (making something up!), so I don't think I would make the tests on particular values more stringent than that?

schlafly:
Looks good to me, merging!

PR #26, opened 2022-12-05
Add PSF module tests.

schlafly:
Move data downloading to CI yaml file.
Add PSF module tests.


schlafly:
This PR adds unit tests for the PSF module.

The challenge here is dealing with the WEBBPSF_PATH and associated data.  I initially tried to get the files downloaded as part of the ci.yaml workflow process.  There may be a good way to do this but I eventually gave up and downloaded the data files as part of the startup process for pytest in conftest.py.  That seemed like a good path but it gets called for each of the different test environments, which means some ~30 times in fairly rapid succession.  The box STScI hosting of the webbpsf data files seems to detect this bad behavior and eventually refuse to serve the files, which generates failing tests.  I now package the 50 MB minimal webbpsf data files in romanisim and grab them directly from there.  That feels like bad form, too, but I'm running out of ideas and want the tests to be reliable.  There will be some maintenance headache as those files and webbpsf are updated, but that's probably going to be the case with the STScI hosting as well.  I don't actually know what the best mechanism is here.

Anyway, I'm satisfied at present that this is a good set of tests for the PSF module, and will defer to the future reducing the maintenance burden of keeping up with the webbpsf releases.

zacharyburnett:
@schlafly I've suggested a few changes that might solve your headache here https://github.com/spacetelescope/romanisim/pull/27

zacharyburnett:
this should probably be squashed when merging (`Squash and Merge`)

PR #27, opened 2022-12-06
hash and cache downloaded data

zacharyburnett:
downloads data and caches it for use in workflow jobs


schlafly:
Thanks Zach.  That looks like a good design.  I like how you have all of the environment variables going into the ci-env.yaml file together to avoid the burden of saving & restoring all of the variables in a number of different places.  It looks like you've sensibly based off of the test-webbpsf branch, so I'll plan on merging that before merging this one.

Thanks for doing this.  I'm not sure I would have ever been able to put all of these pieces together!

zacharyburnett:
> Thanks Zach. That looks like a good design. I like how you have all of the environment variables going into the ci-env.yaml file together to avoid the burden of saving & restoring all of the variables in a number of different places. It looks like you've sensibly based off of the test-webbpsf branch, so I'll plan on merging that before merging this one.
> 
> Thanks for doing this. I'm not sure I would have ever been able to put all of these pieces together!

no problem! I just need to fix the home directory on MacOS and then it will be ready

PR #29, opened 2022-12-20
Add unit tests for image module

schlafly:
This PR adds unit tests for the image module.  It also modestly refactors the image module in order to make the unit tests more manageable.
It also adds a couple largely unrelated features:
* GALSIM_CAT_PATH now falls back to the default so that if you have these catalogs installed normally you don't have to do anything special to get tests running.
* romanisim-make-image now takes an --nobj argument so that fewer than the default number of objects can optionally be simulated.

### DIFF
--- a/romanisim/tests/test_image.py
+++ b/romanisim/tests/test_image.py
@@ -1,0 +1,261 @@
+"""Unit tests for image module.
+
+These routines exercise the following:
+- make_l2: resultants -> l2, eventually belongs in an l2 module
+- in_bounds: check if points are in bounds
+- add_objects_to_image: Adds some objects to an image.
+- simulate_counts_generic: Adds sky, dark, and an object list to an image.
+- simulate_counts: Wraps simulate_counts_generic, making sky & dark images
+  for Roman to pass on 
+- simulate: reads in Roman calibration files.  Sends to simulate_counts, and
+  further passes that to L1 and L2 routines.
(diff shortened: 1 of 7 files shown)


schlafly:
Okay, this looks good to review.  @PaulHuwe , any interest?  The changes appear more extensive than one would like, but it's mostly refactoring much of the body of simulate_counts into a pair of additional functions that don't need to know about CRDS, etc..  Factoring them out makes it easier to test them.  image.py and l1.py were the last remaining large-ish modules without good tests, so this starts making the test coverage look good.  I'll want to chat about regression tests and SOC tests some time soon---probably in the new year, though!

romanisim/tests/test_image.py
```diff
@@ -0,0 +1,258 @@
...
+    objinfo = image.simulate_counts_generic(
+        im, exptime, objlist=imdict['chromcatalog'],
+        psf=imdict['impsfchromatic'], xpos=[1000, 1000],
+        ypos=[1000, 1000], zpflux=zpflux)
+    assert np.sum(objinfo['counts'] > 0) == 0
+    # these sources should be out of bounds
+
+
+def test_simulate_counts():
+    # def simulate_counts(sca, targ_pos, date, objlist, filter_name,
```
PaulHuwe, line 181: Did you mean to delete this block?

PaulHuwe (commented):
LGTM.

PR #30, opened 2022-12-20
Add unit tests for L1 module

schlafly:
This PR adds unit tests for the L1 module.  This module is pretty well self-contained and testing works out without any surprises.  There were a couple of minor improvements as well, allowing the L1 module to simulate images that aren't 4k x 4k (to speed tests), and to optionally not use gain measurements.  A few unused vestigial arguments were also removed.

I could not think of especially good tests of correctness.  Even reasonable things like "ramps increase" or "ramps don't measure more photons in total than were accumulated in the whole exposure" aren't true because of read noise.  I turn off read noise and do some basic tests, and then turn it back on and just do sanity tests (... is the shape right?), as well as testing that the read noise applied looks like what one would expect.

It would be virtuous to check that the correlations between reads are what one would expect given the Poisson noise covariances.  I have played that game before but am not presently playing it as part of the unit tests.  I could consider adding that in a later PR but don't want to focus on it now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           fetch-depth: 0
       - run: echo "DATA_PATH=$HOME/work/romanisim/romanisim/romanisim/data" >> $GITHUB_ENV
       - run: |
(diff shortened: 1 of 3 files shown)


schlafly:
@PaulHuwe , last one of these for a while I hope; any interest reviewing?

PaulHuwe (approved):
LGTM

PR #32, opened 2023-01-10
Add SOC tests for DMS 214, 215, 216 (image rendering)

schlafly:
This adds tests demonstrating PSF rendering, analytic model galaxy rendering, and L2 image creation.

This decorates an existing test that made an image in L2 format, and additionally makes some new image rendering tests, intending to demonstrate that we successfully put PSFs and galaxies in images.  I have tried to make these tests non-trivial by avoiding using galsim to do the rendering, and instead doing my own dumb brute force rendering techniques---basically just rendering the images in a hugely oversampled way and then binning down.  I then verify that the inner few arcseconds of the resulting images agree at the 5 sigma level, including only Poisson errors, for sources with 100k total counts. Agreement is quite good, though there are small differences everywhere due to the imperfect integration over pixels and different treatment of the extreme wings of the PSF where there is no flux.

The tests identified a few issues which were also addressed:
- we formerly did object.withFlux(...) on the convolved object, and now do it on the source profile.  The difference is that the PSF profile may not sum to 1 exactly, as it contains only an image of the PSF in a finite aperture.  This leads to a change in the flux scale of ~2%.
- Before we were unintentionally using a convention where pixel 1 is the center of the first pixel; now pixel 0 is the center of the first pixel.  This now matches the pixel_to_world convention used by GWCS and inherited here.
- I previously believed that webbpsf contained distortion in its modeling, but it does not.  This simplifies construction of the PSF slightly.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -147,6 +147,9 @@ def add_objects_to_image(image, objlist, xpos, ypos, psf,
                          rng=None, seed=None):
     """Add sources to an image.
 
+    Note: this includes Poisson noise when photon shooting is used
+    (i.e., for chromatic source profiles), and otherwise is noise free.
+
     Parameters
     ----------
     image : galsim.Image
@@ -202,17 +205,19 @@ def add_objects_to_image(image, objlist, xpos, ypos, psf,
     for i, obj in enumerate(objlist):
         t0 = time.time()
         image_pos = galsim.PositionD(xpos[i], ypos[i])
-        final = galsim.Convolve(obj.profile * flux_to_counts_factor, psf)
+        profile = obj.profile
+        if not chromatic:
+            if obj.flux is None:
+                raise ValueError('Non-crhomatic sources must have specified '
+                                 'fluxes!')
+            profile = profile.withFlux(
+                obj.flux[filter_name] * flux_to_counts_factor)
+        final = galsim.Convolve(profile * flux_to_counts_factor, psf)
         if chromatic:
             stamp = final.drawImage(
                 bandpass, center=image_pos, wcs=image.wcs.local(image_pos),
                 method='phot', rng=rng)
         else:
-            if obj.flux is None:
-                raise ValueError('Non-chromatic sources must have specified '
-                                 'fluxes!')
-            final = final.withFlux(
-                obj.flux[filter_name] * flux_to_counts_factor)
             try:
                 stamp = final.drawImage(center=image_pos,
                                         wcs=image.wcs.local(image_pos))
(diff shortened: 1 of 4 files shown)


schlafly:
@ddavis-stsci , @tddesjardins , any interest in reviewing this?  I added three tests intended to demonstrate that the simulator can:
- add point sources to images (DMS 214)
- add galaxies to images (DMS 215)
- construct L2 images (DMS 216)
I realize that we'll have more formal verification of these requirements at a later stage, but I wanted to know if you think these tests address the spirit of those requirements in a useful way.  I'm presently writing these as unit tests rather than regression tests; if the latter make more sense, please let me know.

romanisim/image.py
```diff
@@ -202,17 +205,19 @@ def add_objects_to_image(image, objlist, xpos, ypos, psf,
     for i, obj in enumerate(objlist):
         t0 = time.time()
         image_pos = galsim.PositionD(xpos[i], ypos[i])
-        final = galsim.Convolve(obj.profile * flux_to_counts_factor, psf)
+        profile = obj.profile
+        if not chromatic:
+            if obj.flux is None:
+                raise ValueError('Non-crhomatic sources must have specified '
```
ddavis-stsci, line 211: step.log.info('DMS79 MSG: Was proper flat fielded ' --> step.log.info('DMS79 MSG: Was proper flat fielded '

ddavis-stsci (commented):
Over all it looks good. In general for the SOC tests we include something like,
step.log.info('DMS79 MSG: Was proper flat fielded '
So that the DMS number will appear in the log messages and the message addresses the requirement.

The one thing that is nagging at me is that these tests seem to determine the internal consistency or that the two different methods of generating the source yield similar results to within a tolerance. 
If the simulated image changes drastically between two runs I don't see how these tests would detect that type of change which I think the regression tests should be able to flag. 
Maybe I'm missing something here?

schlafly:
Thanks, I'll add that line.

Yes, I had seen it as a feature that when webbpsf changes something about the optics modeling for Roman that tests will still pass.  I could imagine adding a measurement of a second moment or something, so that if the second moment is too different from the current second moment then tests fail; would that address your concern?

ddavis-stsci:
I think I understand what you are wanting to do, allow benign or 
beneficial updates to the libraries to be
not flagged as errors.

What I don't understand is how the code can distinguish between the 
changes you want to allow and any
changes that result from bugs or errors in the libraries.

Maybe this is the wrong venue to discuss this. The real question is if 
Megan and the testing group think
that these tests verify the requirements. If so I'm good with it but we 
might want to check.

schlafly:
@tddesjardins , I've merged this for now, and will spin out any comments into a separate ticket---thanks!

PR #33, opened 2023-01-19
add regtests for Jenkins job

zacharyburnett:
https://plwishmaster.stsci.edu:8081/job/RT/job/romanisim/


zacharyburnett:
the regtests have the following error:
```
E   ModuleNotFoundError: No module named 'asdf.fits_embed'
```

WilliamJamieson:
> the regtests have the following error:
> 
> ```
> E   ModuleNotFoundError: No module named 'asdf.fits_embed'
> ```

I assume this is attempting to pull from the asdf main branch? If so, that module has been deprecated and removed.

romanisim/regtest/regtestdata.py
```diff
@@ -0,0 +1,490 @@
...
+    # Figure out whether we have a config or class
+    step = step_params['step']
+    if step.endswith(('.asdf', '.cfg')):
+        step = os.path.join('config', step)
+
+    # Run the step
+    full_args = [step, rtdata.input]
+    full_args.extend(step_params['args'])
+
+    RomanStep.from_cmdline(full_args)
```
zacharyburnett, line 325: I'm not sure how to port over this part of the Roman tests; would the simulator run here?

schlafly: I thought I understood that @nden and you decided that the romancal infrastructure might not be the right template for romanisim in the first place?  She pointed me to 
https://github.com/spacetelescope/simcal/blob/main/simcal/jwst_tests/imaging/test_nircam.py
as an alternative, more stripped-down approach to interacting with stored regression test data.  Has that thinking evolved?

I agree that that code looks pretty attached to the romancal / stpipe 'Step' infrastructure which we don't have in romanisim; I don't immediately know what the equivalent would be.  Maybe we should huddle to touch base about this some time soon?

schlafly: FWIW, just having glanced through that once, I had the impression that things looked lightweight enough that I would be able to put things together---the main new bit for me was going to be to learn how to put files in artifactory.  I was planning to take a swing at this soon and then call for help if I got stuck again.  But you know better than I do how much stuff is going on behind the scenes.

JenkinsfileRT_dev
```diff
@@ -0,0 +1,56 @@
...
+        stage('Test') {
+            steps {
+                sh("mkdir data")
+                sh("mkdir data/galsim-data")
+                sh("curl https://github.com/GalSim-developers/GalSim/raw/releases/2.4/examples/data/real_galaxy_catalog_23.5_example.fits -o data/galsim-data/real_galaxy_catalog_23.5_example.fits")      
+                sh("curl https://github.com/GalSim-developers/GalSim/raw/releases/2.4/examples/data/real_galaxy_catalog_23.5_example_selection.fits -o data/galsim-data/real_galaxy_catalog_23.5_example_selection.fits")      
+                sh("curl https://github.com/GalSim-developers/GalSim/raw/releases/2.4/examples/data/real_galaxy_catalog_23.5_example_fits.fits -o data/galsim-data/real_galaxy_catalog_23.5_example_fits.fits")      
+                sh("""#!/usr/bin/env bash
+                      source $WORKSPACE/miniconda3/etc/profile.d/conda.sh
+                      export WEBBPSF_PATH="/grp/jwst/ote/webbpsf-data"
```
schlafly, line 44: I am slightly worried about this; I don't control this, and when it is updated, regression tests will start to fail?

zacharyburnett: ok, I can try reverting to a different download URL and see if that works

requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 git+https://github.com/asdf-format/asdf
-git+https://github.com/astropy/astropy
 git+https://github.com/spacetelescope/roman_datamodels
+astropy>=0.0.dev0
+git+https://github.com/spacetelescope/stpipe
```
schlafly, line 4: We do have a real astropy dependency and I understand making it part of dev requirements.  I don't think we have any stpipe dependency in the simulator and think we could leave that line out.

schlafly (approved):
Looks good to me.  One minor question re the dev dependencies, but good to merge.

requirements-dev.txt
zacharyburnett: good catch, fixed

zacharyburnett:
@schlafly I've been dumb this whole time, I didn't use `-L` with `curl` so it was just downloading the HTML page and calling it an archive; that's why the data was empty. Fixed run is currently running: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2Fromanisim/detail/romanisim/80/pipeline/

PR #34, opened 2023-01-24
Changes supporting adding tests for uncalibrated image simulation

schlafly:
- Add function trimming input catalogs to sources vaguely near the detector
- Allow sending either list[CatalogObject] or Table to simulate_counts / simulate_counts_generic
- Make code more efficient when sending large number of sources, most of which fall off the detector.
- Refactor IPC code to improve testing.
- Make default ra/dec specification specify the location of the center of the WFI array rather than the boresight.
  - Update WCS tests & documentation accordingly.
- Add SOC tests for requirements:
  - DMS220: Adding read noise.
  - DMS225: inject a source into a ramp.
  - DMS226: Convolving an image with an IPC kernel.
  - DMS227: Making an L1 output file with the right structure.

### DIFF
--- a/romanisim/catalog.py
+++ b/romanisim/catalog.py
@@ -15,7 +15,11 @@ from . import util
 
 @dataclasses.dataclass
 class CatalogObject:
-    """Simple class to hold galsim positions and profiles of objects."""
+    """Simple class to hold galsim positions and profiles of objects.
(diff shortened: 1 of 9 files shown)


PaulHuwe (approved):
LGTM

PR #35, opened 2023-01-30
Simple astrometric converter to take a Gaia catalog and make a romani…

schlafly:
…sim input catalog at a particular epoch.

### DIFF
--- a/romanisim/gaia.py
+++ b/romanisim/gaia.py
@@ -1,0 +1,74 @@
+import numpy as np
+from astropy import table, coordinates, time, units as u
+
+
+def gaia2romanisimcat(gaiacat, date, refepoch=2016.0, boost_parallax=1,
+                      fluxfields=['F184']):
+    """Convert Gaia output to a catalog of locations for input to romanisim.
+
+    Mostly an astrometry routine, doing the usual parallax and proper motion
+    computations by hand.  Caveat emptor---may all be wrong!
+
+    boost_parallax is intended to allow accounting for the fact that parallaxes
+    are for 1 AU, but Roman will be at 1.01 AU (L2) and so everything moves by
+    1% more than you might expect.  If one wants to do better than that one
+    needs to think about the detailed orbit of Roman in L2; the location
+    of L2 itselfs moves annually and monthly due to the eccentricity of the
+    Earth's orbit and the fact that the Earth is not at the Earth-moon
+    barycenter.
+
+    Parameters
+    ----------
+    gaiacat : astropy.table.Table
+        Gaia catalog, containing at least ra, dec, pmra, pmdec, parallax,
+        phot_g_mean_mag fields
+    date : astropy.time.Time
+        Time of observation
+    refepoch : float
+        Reference epoch of Gaia parameters
+    boost_parallax : float
+        Amount to boost parallaxes by (L2 describes a wider orbit than the
+        Earth's)
+    fluxfields : list[str]
+        List of strings to fill with estimated fluxes
+
+    Returns
+    -------
+    catalog : astropy.table.Table
+        astropy Table formatted for input to romanisim
+    """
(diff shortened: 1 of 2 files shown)


schlafly:
@mairanteodoro , any interest in reviewing this PR?  I just generated a bunch of images using it which seem to have worked?
Steps:
- download Gaia VO table from Gaia archive
- convert to input file romanisim-test-catalog-2026.0.ecsv using this PR
- Run simulations and stpipe on the resulting images
```
seq 1 18 | parallel -j 8 romanisim-make-image --catalog ./romanisim-test-catalog-2026.0.ecsv --radec 270.0 66.0 --bandpass F158 --sca {} --usecrds --webbpsf --date 2026 1 1 --level 1 l1-270-66-gaia-2026-sca{}.asdf
seq 1 18 | parallel -j 8 strun --disable-crds-steppars --steps.jump.rejection_threshold=180. --steps.jump.three_group_rejection_threshold=185. --steps.jump.four_group_rejection_threshold=190 roman_elp l1-270-66-gaia-2026-sca{}.asdf
```
That seems to have worked.

The most interesting content for review would be things like whether I implemented proper motion & parallax correctly and got the units and signs and factors of two or cos(theta) right.  If you have background there, feel free to take a look.  Otherwise I don't feel that bad about this code.

mairanteodoro:
@schlafly Are you going to address the failing tests before merging? Or are they irrelevant for now?

schlafly:
This is due to the unit changes rather than something about the PR.  I'll probably add in the unit stuff early next week.

romanisim/gaia.py
```diff
@@ -0,0 +1,73 @@
...
+    fluxfields : list[str]
+        List of strings to fill with estimated fluxes
+
+    Returns
+    -------
+    catalog : astropy.table.Table
+        astropy Table formatted for input to romanisim
+    """
+    outcat = table.Table()
+    dt = date.jyear - refepoch
```
mairanteodoro, line 41: For the sake of consistency, I think we should define `refepoch` as a `Time` object (like `date`) and do `dt = date.jyear - refepoch.jyear`.

romanisim/tests/test_gaia.py
```diff
@@ -0,0 +1,40 @@
...
+    fakegaiacat['ra'] = np.array([270.0] * 5) * u.deg
+    fakegaiacat['dec'] = np.array([66.0, 66.0, -24.0, 0, 0]) * u.deg
+    fakegaiacat['pmra'] = np.array([1, 0, 0, 1, 0]) * u.mas / u.year
+    fakegaiacat['pmdec'] = np.array([1, 0, 0, 1, 0]) * u.mas / u.year
+    fakegaiacat['parallax'] = np.array([0, 1, 1, 0, 0]) * u.mas
+    fakegaiacat['phot_g_mean_mag'] = 15.0
+    dates = Time(
+        np.linspace(2016, 2017, 10), format='jyear')
+    cats = [gaia.gaia2romanisimcat(
+        fakegaiacat, date, refepoch=2016, boost_parallax=1) for date in dates]
```
mairanteodoro, line 20: `[...], refepoch=Time(2016, format='jyear'), [...]`

mairanteodoro (approved):
Looks good to me.

romanisim/gaia.py
schlafly: I thought about that.  But I don't think as a user we ever really want to have more flexibility than accepting Julian years for this routine.  If I had the more general Time interface, that would let people use whatever random times they have floating around, which I guess would be nice.  But it would also mean that when Gaia says their reference epoch is 2016.0, I'd need to read about how to transform 2016.0 into an astropy Time object, rather than just entering 2016 here.  I was happy to see that Time(..., format='jyear') was a thing, but I'm happy to save future users needing to relearn that.

PR #36, opened 2023-02-08
Add CRs to simulator

schlafly:
This PR adds @jwuphysics 's CR module to the simulator, adding CRs as reads are simulated in the L1 module.

### DIFF
--- a/docs/romanisim/l1.rst
+++ b/docs/romanisim/l1.rst
@@ -30,5 +30,11 @@ Interpixel capacitance (IPC) is added following non-linearity and before read-ou
 
 This is slightly different than including IPC in the PSF kernel because including IPC in the PSF kernel leaves the Poisson noise uncorrelated.
 
+Cosmic rays
+-----------
+Cosmic rays are added to the simulation read-by-read.  The cosmic ray parameters follow Wu et al. (2023).  The locations of cosmic rays are chosen at random to sample the focal plane uniformly.  Lengths are chosen according to a power law distribution :math:`p(l) \\sim l^{-4.33}`, with lengths between 10 and 10,000 microns.  Charge deposition rates per micron are selected from a Moyal distribution located at 120 electrons per micron with a width of 50 electrons per micron.  An idealized charge is computed for each pixel in a read according to the product of the deposition rate per micron and the length of the cosmic ray's path within that pixel.  This idealized charge is Poisson sampled and added to the relevant pixels in a read.
+
+
 .. automodapi:: romanisim.l1
 .. automodapi:: romanisim.nonlinearity
+.. automodapi:: romanisim.cr
--- a/romanisim/cr.py
+++ b/romanisim/cr.py
@@ -4,11 +4,11 @@ import scipy.interpolate as interpolate
 
 def create_sampler(pdf, x):
     """A function for performing inverse transform sampling.
-    
+
     Parameters
     ----------
     pdf : callable
-        A function or empirical set of tabulated values which can 
+        A function or empirical set of tabulated values which can
         be used to call or evaluate `x`.
     x : 1-d array of floats
         A grid of values where the pdf should be evaluated.
@@ -20,17 +20,17 @@ def create_sampler(pdf, x):
         from the `pdf` distribution within the bounds described
         by the grid `x`.
     """
-    
-    y = pdf(x)                        
-    cdf_y = np.cumsum(y) - y[0]          
-    cdf_y /= cdf_y.max()       
+
+    y = pdf(x)
+    cdf_y = np.cumsum(y) - y[0]
+    cdf_y /= cdf_y.max()
     inverse_cdf = interpolate.interp1d(cdf_y, x)
     return inverse_cdf
 
 
 def moyal_distribution(x, location=1000, scale=300):
-    """Return unnormalized Moyal distribution, which approximates a 
-    Landau distribution and is used to describe the energy loss 
+    """Return unnormalized Moyal distribution, which approximates a
+    Landau distribution and is used to describe the energy loss
     probability distribution of a charged particle through a detector.
 
     Parameters
@@ -48,7 +48,7 @@ def moyal_distribution(x, location=1000, scale=300):
         Moyal distribution (pdf) evaluated on `x` grid of points.
     """
     xs = (x - location) / scale
-    moyal = np.exp(-(xs + np.exp(-xs)) / 2) 
+    moyal = np.exp(-(xs + np.exp(-xs)) / 2)
     return moyal
 
 
@@ -63,7 +63,7 @@ def power_law_distribution(x, slope=-4.33):
     slope : float
         The log-log slope of the distribution, default based on
         Miles et al. (2021).
-    
+
     Returns
     -------
     power_law : 1-d array of floats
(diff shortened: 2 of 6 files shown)


schlafly:
@jwuphysics , if you're up for it, please take a look at the one silly test I added to the test_cr module.  The CRs the module was generated looked a little too disconnected to me, so I think either there is a bug, I've broken something, or I've misunderstood something.  For more context, here's a picture of a CR that I think was intended to be drawn along the line
image
So there's an off by one error in one dimension which is probably irrelevant.  Additionally, the crossing into the 3rd pixel is ~missed, and even though traverse(...) returns three pixels, two of them are identical, albeit with different lengths (see test).  I haven't tried to diagnose what's up here.

jwuphysics:
Thanks @schlafly. Looking into this!

jwuphysics:
Looks like there was an error in the business logic, specifically with the use of `np.floor` when it would be more sensible to use `np.round`. I've re-written the function from the ground up (see below) and it passes all of my quick tests. Thanks for checking this!

```python

def traverse(trail_start, trail_end, N_i=4096, N_j=4096):
    """Given a starting and ending pixel, returns a list of pixel 
    coordinates (ii, jj) and their traversed path lengths. Note that
    the centers of pixels are treated as integers, while the borders
    are treated as half-integers.

    Parameters
    ----------
    trail_start: two-element array of floats
        The starting coordinates in (i, j) of the cosmic ray trail, 
        in units of pix.
    trail_end: two-element array of floats, units of pix
        The ending coordinates in (i, j) of the cosmic ray trail, in 
        units of pix.

    Returns
    -------
    ii : 1-d int array of shape N
        i-axis positions of traversed trail, in units of pix.
    jj : 1-d array of shape N
        j-axis positions of traversed trail, in units of pix.
    lengths : 1-d array of shape N
        Chord lengths for each traversed pixel, in units of pix. 
    """

    # increase in i-direction
    if trail_start[0] < trail_end[0]:
        i0, j0 = trail_start
        i1, j1 = trail_end
    else:
        i1, j1 = trail_start
        i0, j0 = trail_end
    
    di = i1 - i0
    dj = j1 - j0
    slope = dj / di
    sign = np.sign(slope)

    # horizontal border crossings at j = integer + 1/2 
    if dj != 0:
        j_horiz = np.arange(np.round(j0), np.round(j1), sign) + 0.5 * sign
        i_horiz = i0 + (di / dj) * (j_horiz - j0)
        cross_horiz = np.transpose([i_horiz, j_horiz])
    else:
        cross_horiz = np.array([[]])

    # vertical border crossings at i = integer + 1/2
    if di != 0:
        i_vert = np.arange(np.round(i0), np.round(i1), 1) + 0.5
        j_vert = j0 + (dj / di) * (i_vert - i0)
        cross_vert = np.transpose([i_vert, j_vert])
    else:
        cross_vert = np.array([[]])

    # compute center of traversed pixel before each crossing
    ii_horiz, jj_horiz = np.round(cross_horiz - np.array([0, 0.5*np.sign(dj)])).astype(int).T
    ii_vert, jj_vert = np.round(cross_vert - np.array([0.5, 0])).astype(int).T

    # combine crossings and pixel centers
    crossings = np.vstack((cross_horiz, cross_vert))
    ii = np.concatenate((ii_horiz, ii_vert))
    jj = np.concatenate((jj_horiz, jj_vert))

    # sort by i axis
    sorted_by_i = np.argsort(crossings[:, 0])
    crossings = crossings[sorted_by_i]
    ii = ii[sorted_by_i]
    jj = jj[sorted_by_i]

    # remove pixels that go outside detector edge -- may not be necessary
    inside_borders = (
        (crossings[:, 0] > -0.5) & (crossings[:, 0] < (N_i - 0.5)) &
        (crossings[:, 1] > -0.5) & (crossings[:, 1] < (N_j - 0.5))
    )
    crossings = crossings[inside_borders]
    ii = ii[inside_borders]
    jj = jj[inside_borders]

    # add final pixel center
    ii = np.concatenate((ii, [np.round(i1).astype(int)]))
    jj = np.concatenate((jj, [np.round(j1).astype(int)]))
    
    # if no crossings, then it's just the total Euclidean distance
    if len(crossings) == 0:
        lengths = np.linalg.norm([di, dj])
    # otherwise, compute starting, crossing, and ending distances
    else:
        first_length = np.linalg.norm(crossings[0] - np.array(trail_start), keepdims=1)
        middle_lengths = np.linalg.norm(np.diff(crossings, axis=0), axis=1)
        last_length = np.linalg.norm(np.array(trail_end) - crossings[-1], keepdims=1)
        print(first_length, middle_lengths, last_length)
        lengths = np.concatenate([first_length, middle_lengths, last_length])

        # remove 0-length trails
        positive_lengths = lengths > 0
        lengths = lengths[positive_lengths]
        ii = ii[positive_lengths]
        jj = jj[positive_lengths]
    
    return ii, jj, lengths

```

jwuphysics:
Once the tech memo is approved/published, would you like me to submit this as a PR? It may still be several weeks away.

schlafly:
Thanks John.  My preference is for you to just directly push to this branch---you have write to the repo, so that should be fine?---though I can copy/paste & push as well.  I can delay merging into main until you're ready.

jwuphysics:
Sounds good to me!

jwuphysics:
Hey I promise I'll push something soon! Another point that I'm aiming to nail down soon is the conversion factor from energy loss to electrons deposited on the detector.

schlafly:
Thanks!  I had to base the persistence PR off this one, since they use the same mechanism for adding instrumental electrons separate from the scene electrons.  Tests were failing there for that reason, so I pushed this and a bit.  I made one more change to subtract 0.5 from the starting locations of the CRs; I think we were inconsistently using center vs edge conventions.

jwuphysics:
Ah yes, you're absolutely correct about that -- I had made that update but not in the snippet I shared with you. I'm going to push something in a moment, but **please consider it a work in progress**. Of course I'd be thrilled if more bugs are found, and tests failed, but I'd like to avoid merging the PR until the tech memo is submitted.

The PR also includes some updated terminology and default values, including:
- Moyal distribution peaks at 120 eV/micron and has a width of 50 eV/micron, which now roughly agrees with the hit statistics on early JWST data
- The parameter `gain` has been renamed `conversion_factor`, which is more appropriate; it also takes on the default value of 0.5 eV/count (i.e. the bandgap energy).
- The detector area has been corrected to 16.8 cm^2.

I don't know how the CI/CD is set up but let me know if I need to install any flake8 hooks or whateever.

jwuphysics:
Hi @schlafly, I've updated the code quite a bit and tried my best to conform to the linting. I think it's ready for another review!

schlafly:
Thanks for this!  This looks good to me.  Just staring at images, the CRs look closer to how I would expect them to look.  We could discuss what the unit tests for this module should look like; they clearly have some value.  I could clearly flesh out the traverse unit test, and maybe add some sanity tests to the simulate_cr routine.  I'm having trouble motivating myself to write tests of the various sampling & stats routines, though.

I've hung the persistence PR off of this one to make use of the same hooks for adding instrumental electrons that aren't scene electrons.  So all things being equal, I'd like to merge this rather than let it get stale.  That obviously doesn't preclude further updates.  If you wanted to tie something to tech memo, another option would be delaying the next romanisim release to after publication.  Relatedly, I will hold off on the bits of the unit tests that claim particular requirements have been satisfied until after you give the go-ahead.

schlafly:
This looks good to go to me.  @PaulHuwe , @jwuphysics , care to give it a once-over?  John, minor doc updates & tests since your last commits; feel free to just give it a LGTM.

jwuphysics (approved):
LGTM!

If I think of any more tests I'll submit a PR, but this looks good.

PR #37, opened 2023-02-21
Add persistence to simulator

schlafly:
This PR adds persistence to the simulator following the Fermi prescription used in Galsim and in Sanchez+2023.

### DIFF
--- a/romanisim/tests/test_persistence.py
+++ b/romanisim/tests/test_persistence.py
@@ -1,6 +1,7 @@
 """Unit tests for persistence module."""
 
 import numpy as np
+import asdf
 from romanisim import persistence
 from romanisim import parameters
 
@@ -28,12 +29,12 @@ def test_fermi():
         rate2 = persist.current((answer[1]) / 60 / 60 / 24)
         assert np.isclose(rate, rate2)
     parameters.persistence['half_well'] = hw
-    hwpersistence = persistence.fermi(hw, 1000, 1, hw, 1, 1, 1)
-    linpersistence = persistence.fermi(hw / 2, 1000, 1, hw, 1, 1, 1)
+    hwpersistence = persistence.fermi(hw, 1000, 1, hw, 100, 1, 1)
+    linpersistence = persistence.fermi(hw / 2, 1000, 1, hw, 100, 1, 1)
     assert np.isclose(hwpersistence / 2, linpersistence)
 
 
-def test_persistence():
+def test_persistence(tmp_path):
     persist = persistence.Persistence()
     img = np.zeros((100, 100), dtype='i4')
     img2 = img.copy()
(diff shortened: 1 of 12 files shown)


schlafly:
@fjaviersanchez , any interest in reviewing this PR?  The diff is a little annoying since I built this off of the CR PR, since that work also needed a mechanism for non-scene photons to each read.  You can ignore the CR component.

Short of finding bugs, perhaps the most useful input would be whether there are good tests to add to the unit test module.  I made up some by-hand things that were intended to test that the Fermi calculation was working as expected---though in retrospect I should probably force the Fermi model parameters to be the ones that those calculations expect.  Then there is some basic testing of the class that is supposed to track the different information about hit pixels.  But maybe there are some additional tests I'm forgetting.

fjaviersanchez:
Thanks @schlafly! Happy to review :).

romanisim/cr.py
```diff
@@ -0,0 +1,326 @@
...
+    for i0, j0, i1, j1, counts_per_pix in zip(
+            cr_i0, cr_j0, cr_i1, cr_j1, cr_counts_per_pix):
+        ii, jj, length_2d = traverse([i0, j0], [i1, j1])
+        length_3d = ((pixel_depth / pixel_size) ** 2 + length_2d ** 2) ** 0.5
+        image[ii, jj] += rng.poisson(counts_per_pix * length_3d)
+
+    return image
+
+
+if __name__ == "__main__":
```
fjaviersanchez, line 316: Not sure if having `__main__` here is intended, but if it is, it would be good to have the ability to change the hardcoded values below at runtime

romanisim/tests/test_persistence.py
```diff
@@ -0,0 +1,51 @@
...
+    answers = [(1, 1000, 1, 1, 1, 1, 1, 0.5),
+               (100, 1000, 1, 1, 1, 1, 1, 100),
+               (100, 1000, 2, 1, 1, 1, 1, 200),
+               (100, 1000, 1, 1, 1, 2, 1, 10000),
+               (100, 1000, 1, 100, 1, 1, 1, 0.5),
+               (100, 1000, 1, 1, 1, 1, 2, 100),
+               (100, 2000, 1, 1, 1, 1, 1, 50),
+               (100, 2000, 1, 1, 1, 1, 2, 25)]
+    hw = parameters.persistence['half_well']
+    parameters.persistence['half_well'] = 0.1
```
fjaviersanchez, line 19: If I understood the test correctly, you never check the case where x is less than the half-well level. If so, I would recommend to add one more line where you check that one.

fjaviersanchez (approved):
The code and tests look good to me. I have a couple of minor comments, but I think this is good to go on my end. Thank you!

romanisim/cr.py
schlafly: John had these in to have some simple tests for the CR module.  I can remove those after we build out the CR tests; let's not worry about that here.

romanisim/tests/test_persistence.py
schlafly: I've added some more here, thanks.

romanisim/cr.py
fjaviersanchez: Got it. Thanks!

romanisim/tests/test_persistence.py
fjaviersanchez: Thanks! Looks good.

PR #38, opened 2023-03-09
`tomli` is implemented read-only as `tomllib` in Python 3.11

zacharyburnett:
`tomli` is implemented read-only as `tomllib` in Python 3.11


schlafly:
I am happy to merge this when you're satisfied.

PR #39, opened 2023-03-15
Add more metadata to the simulation output headers.

schlafly:
This PR fills out a bit more of the simulation metadata to make it easier for users to find things.

It also fixes one bug where Poisson noise was being applied twice.

### DIFF
--- a/romanisim/catalog.py
+++ b/romanisim/catalog.py
@@ -226,7 +226,8 @@ def table_to_catalog(table, bandpasses):
         elif table['type'][i] == 'SER':
             obj = galsim.Sersic(table['n'][i], table['half_light_radius'][i])
             obj = obj.shear(
-                q=table['ba'][i], beta=(table['pa'][i] + np.pi / 2) * galsim.radians)
+                q=table['ba'][i],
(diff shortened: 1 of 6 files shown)


schlafly:
@PaulHuwe, have you seen this kind of fits_embed failure before?  I'm embarrassed that I don't actually understand test-xdist vs. test-devdeps.

PaulHuwe:
Yes - devdops is testing against the latest dev versions of dependencies listed in `requirements-dev.txt` - which has some significant changes. (I have run into these and started changing code for them in implementing RDM support.) That's why you are getting these errors.

schlafly:
Should I ignore these for the purposes of deciding whether to merge?

schlafly:
@PaulHuwe , is it okay for me to merge this?  Unfortunately it does touch the metadata and will probably cause some merge headaches.

PaulHuwe:
Yes, I will fix them all in my PR.

PaulHuwe (approved):
LGTM with the proviso that I will be modifying this a bit in my PR.

PR #40, opened 2023-03-21
Improve API for creating parametric distribution of sources for simulation

schlafly:
This PR exposes more of the API for creating parametric distributions of sources.  Different magnitude distributions can be selected via the power law index dlog10 N / dmag, and different spatial distributions are possible (uniform in disc; King profiles).  I've also included some associated tests and documentation.

It is expected that most scientific applications will not use these routines, which are simplistic, but this allows easy access to some vaguely appropriate toy simulations.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -16,7 +16,7 @@ import galsim
 from galsim import roman
 try:
     import roman_datamodels.testing.utils as maker_utils
-except:
+except ImportError:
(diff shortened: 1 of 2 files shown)


schlafly:
@PaulHuwe , any interest in reviewing this?  Slight extensions to the catalog generation routines to try to make them flexible enough for me to imagine that they can be used to "generate parametric source distributions."

PR #41, opened 2023-03-22
use OpenAstronomy workflows and move MacOS jobs to weekly schedule

zacharyburnett:
This PR abstracts CI workflows to the OpenAstronomy workflows; this should reduce maintenance for updating and maintaining actions. Additionally, these changes move the majority of MacOS jobs to the weekly scheduled workflow. This should alleviate the issue where the limited MacOS runners for the organization are not available for CI jobs.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,7 @@ jobs:
           echo "galsim_url=https://github.com/GalSim-developers/GalSim/raw/releases/2.4/examples/data/" >> $GITHUB_OUTPUT
           echo "path=/tmp/data" >> $GITHUB_OUTPUT
       - run: |
+          mkdir ${{ steps.data.outputs.path }}
           wget ${{ steps.data.outputs.webbpsf_url }} -O ${{ steps.data.outputs.path }}/minimal-webbpsf-data.tar.gz
           cd ${{ steps.data.outputs.path }}
           tar -xzvf minimal-webbpsf-data.tar.gz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
         WEBBPSF_PATH: ${{ needs.data.outputs.webbpsf_path }}
         GALSIM_CAT_PATH: ${{ needs.data.outputs.galsim_cat_path }}
       cache-path: ${{ needs.data.outputs.path }}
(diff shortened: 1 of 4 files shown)


schlafly:
Thanks Zach.  What do you recommend here?  There are two issues:
- an xml issue
- a concatenate issue

Neither of these are in direct romanisim imports.  Instead in the first case romanisim is importing webbpsf is importing matplotlib is importing is importing PIL is importing defusedxml, where there is apparently a dependency problem?  I would have imagined that this means something like defusedxml has an incorrect python version requirement.  In the second case it looks like it's an astropy / numpy thing; all I'm doing is accessing a property of an astropy object.

I don't really understand the "with deprecation" bit, but I don't really understand the difference between the new and old workflows.

zacharyburnett:
> Thanks Zach. What do you recommend here?

It looks like `defusedxml==0.7.1` works, but `0.3` does not. It's unclear to me when the `_IterParseIterator` issue was actually fixed, so I will experiment with versions until it works.  

> I don't really understand the "with deprecation" bit, but I don't really understand the difference between the new and old workflows.

That was my assumption, as these pieces of the API are no longer available in `defusedxml`, but it looks now like that's an issue that was resolved in later releases.

schlafly:
But, sorry, trying to understand better---do you think that this defusedxml bug is a bug in romanisim's requirements or in ~matplotlib's requirements?  For most of these failures romanisim is just importing matplotlib to trigger the bug, which to me means that we should be filing an issue on matplotlib if we wanted this resolved?  I mean, we clearly can specify newer versions that do work, and that also seems appropriate to help users, but I feel like at some level we're not responsible for this issue.

zacharyburnett:
> But, sorry, trying to understand better---do you think that this defusedxml bug is a bug in romanisim's requirements or in ~matplotlib's requirements? For most of these failures romanisim is just importing matplotlib to trigger the bug, which to me means that we should be filing an issue on matplotlib if we wanted this resolved? I mean, we clearly can specify newer versions that do work, and that also seems appropriate to help users, but I feel like at some level we're not responsible for this issue.

yes that's a good point; we can pin the minimum dependency for now, but it should be `matplotlib` that fixes the issue if all we're doing is importing it to trigger the bug

schlafly (approved):
LGTM, thanks!

PR #43, opened 2023-03-31
RCAL-388 Integrate RDM Usage to RomanISim

PaulHuwe:
This is an attempt to introduce roman_datamodel compliant usage to the simulator.


schlafly:
Is romancal now actually a dependency?  It looks like there's an issue there in the tests of old versions that I am not immediately able to parse.

PR #46, opened 2023-04-12
Remove romancal dependency.

schlafly:
This removes the simulator's dependency on romancal, which should not be needed and is causing test failures.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,6 @@ dependencies = [
     'galsim >=2.3',
     'rad >=0.14.1',
     'roman_datamodels >=0.14.1',
-    'romancal >=0.10',
     'gwcs >=0.18.1',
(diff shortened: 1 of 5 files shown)


WilliamJamieson:
Romancal isn't causing the problem it's the codecov dependency.

schlafly:
@PaulHuwe , any interest in reviewing this?  I originally wanted to get rid of the romancal dependency since it was triggering test failures in the -oldestdeps tests.  We didn't have any explicit dependencies on romancal, but we did on stpipe through stpipe.crds_client.  I didn't immediately see the improvement of stpipe.crds_client over crds itself, so I changed those.  Then for some reason we had a new unrelated test failure, which I also fixed.

PaulHuwe (approved):
LGTM

PR #51, opened 2023-05-09
drop support for Python 3.8

zacharyburnett:
Numpy has stopped supporting Python 3.8 in new versions


schlafly:
Is it surprising that dropping python 3.8 suppor leads to a test failure on test-oldestdeps-cov-xdist?  It looks to be when the cython module is compiled, so it seems like it could be a real failure uncovered by this issue.

schlafly:
I resolved this issue in https://github.com/spacetelescope/romanisim/pull/54, we can close this one.  Indeed, I needed to bump the version of cython up to use python 3.9.

PR #52, opened 2023-05-09
fix readthedocs configuration

zacharyburnett:
attempts to fix RTD build by using the new `build` block (`python.version` is deprecated) and also building the environment with `mamba` to avoid issues with availability of `openssl` and `graphviz`


schlafly (approved):
Looks good to me, thank you!

PR #56, opened 2023-06-05
use new wheel index URL

zacharyburnett:
Scipy updated their nightly wheel index location


schlafly (approved):
LGTM, feel free to merge when ready.

PR #57, opened 2023-06-14
"Simple" example using romanisim-make-image for docs

bmorris3:
In learning to use `romanisim`, I found that the [Running the simulation](https://romanisim.readthedocs.io/en/latest/romanisim/running.html) page in the docs explains the arguments, but does not have examples.

Users want/will want to have a reproducible usage example in the docs, so I've added one here.

### DIFF
--- a/docs/romanisim/running.rst
+++ b/docs/romanisim/running.rst
@@ -80,3 +80,83 @@ modeling.
 		location of the center of the WFI array or the
 		telescope boresight, when the ``--boresight`` argument
 		is specified.
+
+
+Example
+=======
+
+Let's put the pieces together. Below, we create a synthetic catalog of bright
+stars centered on the center of M13, observed with SCA 1 in the F087 filter.
+Though we create 10,000 stars, only 86 will ultimately fall on SCA 1.
+In this Python example, we synthesize images of those stars with ``romanisim`` via
+the `~romanisim.image.simulate` function, and plot the results
+with ``matplotlib``.
+
+.. code-block:: python
+
+    from copy import deepcopy
+
+    import asdf
+    import numpy as np
+    import matplotlib.pyplot as plt
+    from astropy.coordinates import SkyCoord
+    from astropy.visualization import simple_norm
+    from galsim import UniformDeviate
+
+    from romanisim import persistence, wcs
+    from romanisim.catalog import make_stars
+    from romanisim.image import simulate
+    from romanisim.parameters import default_parameters_dictionary
+
+    # use the coordinate of M13 as the center of the detector:
+    coord = SkyCoord.from_name("M13")
+
+    # choose an SCA to simulate:
+    sca = 1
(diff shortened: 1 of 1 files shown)


schlafly:
Looks good to me; feel free to merge you're ready.  For this particular case one could imagine calling romanisim.image.simulate(...) directly, without the romanisim-make-image wrapper, but this is a nice approach as well, and doubles as a demo of the CLI.

bmorris3:
The motivation for using the CLI version in the above script instead of `simulate`, was really that there are a bunch of tasks that the CLI version takes care of for you, which I wanted to avoid duplicating:
https://github.com/spacetelescope/romanisim/blob/a6a8ecd6964e1534ef6419b95eaa4868edb5f06a/scripts/romanisim-make-image#L62-L92
It may be worthwhile to prevent users from kicking the can down the road on learning these necessary inputs though, especially for this doc example. I'll edit with `simulate`!

bmorris3:
@schlafly This updated version uses `simulate` rather than the unnecessary `check_call`.

docs/romanisim/running.rst
```diff
@@ -80,3 +80,87 @@ modeling.
...
+    # choose an SCA to simulate:
+    sca = 1
+    filt = 'F087'
+    seed = 0
+
+    # save the star catalog here:
+    catalog_path = 'small-synthetic-catalog.ecsv'
+
+    # write out the result to ASDF:
+    output_path = 'small-synthetic-image.asdf'
```
schlafly, line 123: I think this is vestigial?

docs/romanisim/running.rst
```diff
@@ -80,3 +80,87 @@ modeling.
...
+    # generate a stellar catalog:
+    cat = make_stars(
+        coord=coord,
+        n=10_000,
+        radius=0.7,
+        bandpasses=[filt],
+        faintmag=18,
+        rng=UniformDeviate(seed=seed)
+    )
+    cat.write(catalog_path, overwrite=True)
```
schlafly, line 134: I think this is vestigial?

docs/romanisim/running.rst
```diff
@@ -80,3 +80,87 @@ modeling.
...
+    # run the simulation:
+    im, simcatobj = simulate(
+        metadata, cat, webbpsf=True, level=2,
+        persistence=persistence.Persistence(),
+        rng=UniformDeviate(seed), usecrds=False
+    )
+
+    asdf_file = asdf.AsdfFile()
+    romanisimdict = {'simcatobj': simcatobj}
+    asdf_file.tree = {'roman': im, 'romanisim': romanisimdict}
```
schlafly, line 154: I think you could drop this, and show im['data'] below instead of asdf_file.tree['roman']['data'].

schlafly (approved):
Looks good to me.  A few minor things you could drop that I think were part of the previous implementation.

PR #61, opened 2023-06-24
RCAL-576 Initial Apt Support

PaulHuwe:
Creation of a script to make multiple images based on CSV input. The script is to either actually make the simulated files or output a text files with appropriate command lines required to create the files (good for computer clusters).

### DIFF
--- a/scripts/romanisim-make-image
+++ b/scripts/romanisim-make-image
@@ -38,6 +38,8 @@ if __name__ == '__main__':
     parser.add_argument('--level', type=int, default=2,
                         help='1 or 2, for L1 or L2 output')
     parser.add_argument('--ma_table_number', type=int, default=1)
+    parser.add_argument('--roll', type=float, default=0,
+                        help='Roll angle for the instrument.')
     parser.add_argument('--seed', type=int, default=None)
     parser.add_argument('--nobj', type=int, default=1000)
     parser.add_argument('--boresight', action='store_true', default=False,
@@ -62,11 +64,11 @@ if __name__ == '__main__':
     if args.radec is not None:
         coord = SkyCoord(ra=args.radec[0] * u.deg, dec=args.radec[1] * u.deg,
                          frame='icrs')
-        wcs.fill_in_parameters(metadata, coord, boresight=args.boresight)
+        wcs.fill_in_parameters(metadata, coord, boresight=args.boresight, roll_ref=args.roll)
 
     if args.date is not None:
-        metadata['exposure']['start_time'] = Time(
-            datetime.datetime(*args.date)).isot
+        metadata['exposure']['start_time'] = Time(Time(
+            datetime.datetime(*args.date)).isot, format="isot", scale="utc")
 
     date = Time(metadata['exposure']['start_time'], format='isot')
 
--- a/scripts/romanisim-make-image
+++ b/scripts/romanisim-make-image
@@ -1,19 +1,11 @@
 #!/usr/bin/env python
 
 import argparse
-import datetime
-import logging
-from copy import deepcopy
-import asdf
 from astropy.coordinates import SkyCoord
 from astropy import units as u
-from astropy.table import Table
-from astropy.io import fits
-from astropy.time import Time
 import galsim
-from romanisim import catalog, image, wcs, persistence
-from romanisim.parameters import default_parameters_dictionary
-from romanisim import log
+from romanisim import log, wcs, persistence
+import ris_make_utils as ris
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(
@@ -21,33 +13,32 @@ if __name__ == '__main__':
         epilog='EXAMPLE: %(prog)s output_image.fits',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument('filename', type=str, help='output image (fits)')
-    parser.add_argument('--catalog', type=str, help='input catalog (csv)',
-                        default=None)
-    parser.add_argument('--radec', type=float, nargs=2,
-                        help='ra and dec (deg)', default=None)
     parser.add_argument('--bandpass', type=str, help='bandpass to simulate',
                         default='F087')
+    parser.add_argument('--boresight', action='store_true', default=False,
+                        help=('radec specifies location of boresight, not '
+                              'center of WFI.'))
+    parser.add_argument('--catalog', type=str, help='input catalog (csv)',
+                        default=None)
+    parser.add_argument('--date', type=str, default=None,
+                        help=('UTC Date and Time of observation to simulate in ISOT format.'))
+    parser.add_argument('--level', type=int, default=2,
+                        help='1 or 2, for L1 or L2 output')
+    parser.add_argument('--ma_table_number', type=int, default=1)
+    parser.add_argument('--nobj', type=int, default=1000)
+    parser.add_argument('--previous', default=None, type=str,
+                        help=('previous simulated file in chronological order '
+                              'used for persistence modeling.'))
+    parser.add_argument('--radec', type=float, nargs=2,
+                        help='ra and dec (deg)', default=None)
+    parser.add_argument('--rng_seed', type=int, default=None)
+    parser.add_argument('--roll', type=float, default=0,
+                        help='Roll angle for the instrument.')
     parser.add_argument('--sca', type=int, default=7, help='SCA to simulate')
     parser.add_argument('--usecrds', action='store_true',
                         help='Use CRDS for distortion map')
     parser.add_argument('--webbpsf', action='store_true',
                         help='Use webbpsf for PSF')
-    parser.add_argument('--date', type=int, nargs='+', default=None,
-                        help=('Date of observation to simulate: '
-                              'year month day hour minute second microsecond'))
-    parser.add_argument('--level', type=int, default=2,
-                        help='1 or 2, for L1 or L2 output')
-    parser.add_argument('--ma_table_number', type=int, default=1)
-    parser.add_argument('--roll', type=float, default=0,
-                        help='Roll angle for the instrument.')
-    parser.add_argument('--seed', type=int, default=None)
-    parser.add_argument('--nobj', type=int, default=1000)
-    parser.add_argument('--boresight', action='store_true', default=False,
-                        help=('radec specifies location of boresight, not '
-                              'center of WFI.'))
-    parser.add_argument('--previous', default=None, type=str,
-                        help=('previous simulated file in chronological order '
-                              'used for persistence modeling.'))
 
     args = parser.parse_args()
 
@@ -59,49 +50,23 @@ if __name__ == '__main__':
                 "packages like galsim's roman package or STIPS may better "
                 "serve such purposes.")
 
-    metadata = deepcopy(default_parameters_dictionary)
+    metadata = ris.set_metadata(date=args.date, bandpass=args.bandpass, sca=args.sca, ma_table_number=args.ma_table_number)
 
     if args.radec is not None:
         coord = SkyCoord(ra=args.radec[0] * u.deg, dec=args.radec[1] * u.deg,
                          frame='icrs')
         wcs.fill_in_parameters(metadata, coord, boresight=args.boresight, roll_ref=args.roll)
 
-    if args.date is not None:
-        metadata['exposure']['start_time'] = Time(Time(
-            datetime.datetime(*args.date)).isot, format="isot", scale="utc")
+    rng = galsim.UniformDeviate(args.rng_seed)
 
-    date = Time(metadata['exposure']['start_time'], format='isot')
+    # Create catalog
+    cat = ris.create_catalog(catalog_name=args.catalog, bandpasses=[args.bandpass], rng=rng, nobj=args.nobj, usecrds=args.usecrds)
 
-    metadata['instrument']['detector'] = f'WFI{args.sca:02d}'
-    metadata['instrument']['optical_element'] = args.bandpass
-    metadata['exposure']['ma_table_number'] = args.ma_table_number
-
-    rng = galsim.UniformDeviate(args.seed)
-    if args.catalog is None:
-        twcs = wcs.get_wcs(metadata, usecrds=args.usecrds)
-        rd_sca = twcs.toWorld(galsim.PositionD(
-            galsim.roman.n_pix / 2, galsim.roman.n_pix / 2))
-        cat = catalog.make_dummy_table_catalog(
-            rd_sca, bandpasses=[args.bandpass], nobj=args.nobj, rng=rng)
-    else:
-        log.warning('Catalog input will probably not work unless the catalog '
-                    'covers a lot of area or you have thought carefully about '
-                    'the relation between the boresight and the SCA locations.')
-        cat = Table.read(args.catalog)
-
+    # Create persistence object
     if args.previous is not None:
         persist = persistence.Persistence.read(args.previous)
     else:
         persist = persistence.Persistence()
 
-    im, extras = image.simulate(
-        metadata, cat, usecrds=args.usecrds,
-        webbpsf=args.webbpsf, level=args.level, persistence=persist,
-        rng=rng)
-
-    romanisimdict = vars(args)
-    romanisimdict.update(**extras)
-
-    af = asdf.AsdfFile()
-    af.tree = {'roman': im, 'romanisim': romanisimdict}
-    af.write_to(args.filename)
+    # Simulate image and write to file
+    ris.simulate_image_file(args, metadata, cat, rng, persist)
--- a/scripts/romanisim-make-stack
+++ b/scripts/romanisim-make-stack
@@ -7,10 +7,8 @@ from astropy import units as u
 from astropy.table import Table
 from astropy.time import Time
 import galsim
-from romanisim import wcs, persistence
+from romanisim import wcs, persistence, log, parameters
 from romanisim.parameters import default_parameters_dictionary
-from romanisim import log
-from romanisim import parameters
 import ris_make_utils as ris
 
 
@@ -40,21 +38,11 @@ def main():
                         default=None,
                         help='Object catalog file for wcs matching (csv)')
 
-    # Debug argument
-    parser.add_argument("--debug",
-                        action='store_true',
-                        help='Display verbose debug output')
-
     # Date & Time argument
     parser.add_argument('-d', '--date',
                         type=str,
                         default='2026-01-01T00:00:00',
                         help=('UTC Date and Time of observation to simulate in ISOT format.'))
-
-    # Flux factor argument
-    parser.add_argument('-f', '--flux_factor',
-                        type=float,
-                        help='An optional flux factor for generated objects to make them brighter.')
 
     # Level argument
     parser.add_argument('-l', '--level',
@@ -85,7 +73,7 @@ def main():
                         help='Filename prefix for simulated files.')
 
     # Persistence
-    parser.add_argument('-p', '--previous',
+    parser.add_argument('-p', '--persistence',
                         action='store_true',
                         help=(
                             'Enable persistence modeling of previous files for exposures after the first in each SCA.'))
@@ -140,13 +128,12 @@ def main():
     ra_diff = max_ra - min_ra
     dec_diff = max_dec - min_dec
 
-    if args.debug:
-        print(f"max_ra = {max_ra}")
-        print(f"min_ra = {min_ra}")
-        print(f"ra_diff = {ra_diff}")
-        print(f"max_dec = {max_dec}")
-        print(f"min_dec = {min_dec}")
-        print(f"dec_diff = {dec_diff}")
+    log.debug(f"max_ra = {max_ra}"
+              f"min_ra = {min_ra}"
+              f"ra_diff = {ra_diff}"
+              f"max_dec = {max_dec}"
+              f"min_dec = {min_dec}"
+              f"dec_diff = {dec_diff}")
 
     # Create random number distribution
     rng = galsim.UniformDeviate(args.rng_seed)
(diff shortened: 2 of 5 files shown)


scripts/romanisim-make-image
```diff
@@ -62,11 +64,11 @@ if __name__ == '__main__':
...
         coord = SkyCoord(ra=args.radec[0] * u.deg, dec=args.radec[1] * u.deg,
                          frame='icrs')
-        wcs.fill_in_parameters(metadata, coord, boresight=args.boresight)
+        wcs.fill_in_parameters(metadata, coord, boresight=args.boresight, roll_ref=args.roll)
 
     if args.date is not None:
-        metadata['exposure']['start_time'] = Time(
-            datetime.datetime(*args.date)).isot
+        metadata['exposure']['start_time'] = Time(Time(
+            datetime.datetime(*args.date)).isot, format="isot", scale="utc")
```
schlafly, line 71: What was the issue with the previous implementation here?  If the issue is that a we'd rather have a Time object than a string in the metadata, then I think you could just drop the '.isot' rather than doubly wrap?

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+from astropy.table import Table
+from astropy.io import fits
+from astropy.time import Time
+import galsim
+from romanisim import catalog, image, wcs, persistence
+from romanisim.parameters import default_parameters_dictionary
+from romanisim import log
+from romanisim import parameters
+
+NUMBER_OF_DETECTORS = 18
```
schlafly, line 21: Can we move this into romanisim.parameters?

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                        default=1000,
+                        help='Specify number of objects to simulate.')
+
+    # Persistence
+    parser.add_argument('-p', '--previous',
+                        type=str,
+                        metavar='previous.asdf',
+                        nargs=1,
+                        help=(
+                            'Previous simulated file in chronological order used for persistence modeling.'))
```
schlafly, line 78: We likely need to think a little more here.  This script is generating images for multiple SCAs, but when simulating SCA 1 we only want to send the previous SCA 1 image.  And for this case where you are providing a list of images to simulate, I think we want to have the script figure this out and not include any persistence on the first image.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+    # Create script making argument
+    parser.add_argument('-m', "--make_script",
+                        action='store_true',
+                        help='Create list of romanisim calls from csv file (e.g. for cluster usage).')
+
+    # Number of objects to simulate
+    parser.add_argument('-n', '--nobj',
+                        type=int,
+                        default=1000,
+                        help='Specify number of objects to simulate.')
```
schlafly, line 70: For this kind of case where we're making a lot of images, we probably don't want to generate an independent random set of sources on each image.  We want instead to generate a big catalog up front.  We should either force the users to supply this catalog or do a bit more with the catalog generation in the code.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                        metavar='previous.asdf',
+                        nargs=1,
+                        help=(
+                            'Previous simulated file in chronological order used for persistence modeling.'))
+
+    # Random number seed
+    parser.add_argument('-r', '--rng_seed',
+                        type=int,
+                        default=None,
+                        help='Random number seed (int)')
```
schlafly, line 84: I think the romanisim-make-image argument is --seed.  We should make those compatible.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                "performed.  For this reason, use of romanisim for "
+                "preparation of ROSES proposals is not advised.  Other "
+                "packages like galsim's roman package or STIPS may better "
+                "serve such purposes.")
+
+    # Set up metadata
+    metadata = deepcopy(default_parameters_dictionary)
+
+    # If creating a script, create the script file
+    if args.make_script:
```
schlafly, line 119: Maybe for this mode, the argument should be
--script
and defaults to None (don't make a script), and otherwise is the name of the file to make?

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+        for option, value in vars(args).items():
+            if args.debug:
+                print(f'option, value = {option, value}')
+            if option not in ('input_file_name','debug','make_script') and value is not None:
+                options_dct[option] = value
+
+
+
+    with open(args.input_file_name, newline='') as infile:
+        reader = csv.DictReader(infile, delimiter=',', skipinitialspace=True)
```
schlafly, line 133: FWIW, we have an astropy dependency, and so I default to Table.read(infile) here.  That's e.g. how the catalogs are read in.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                        line += f" --date "
+                        for s in date.strftime('%Y %m %d %H %M %S %f').split(' '):
+                            line += f"{int(s)} "
+
+                        # print line
+                        if args.debug:
+                            print(f'line = {line}')
+
+                        # add line to script
+                        spcript_file.write(line+'\n')
```
schlafly, line 201: Two cents: I'd break this section out to a separate function outside the for entry in reader level.  This doesn't use anything else we've really computed above?

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                    # Break the loop if only one detector specified
+                    if args.sca > 0:
+                        break
+                    else:
+                        sca += 1
+
+                # Calculate time offset for the next exposure
+                openshuttertime = parameters.read_time * (
+                        ma_table[-1][0] + ma_table[-1][1] - 1)
+                time_offset += (openshuttertime + (float(entry['overhead_time']))) * u.s
```
schlafly, line 256: Let's use the 'duration' column I proposed here:
https://jira.stsci.edu/browse/APT-93485
So this would just be time_offset += entry['duration'] * u.s, I think.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+
+    with open(args.input_file_name, newline='') as infile:
+        reader = csv.DictReader(infile, delimiter=',', skipinitialspace=True)
+        for entry in reader:
+            # Debug: print keys
+            if args.debug:
+                print(f'entry.keys = {entry.keys()}')
+
+            # Reset time offset
+            time_offset = 0 * u.s
```
schlafly, line 140: This gets called every time and isn't what we want?

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+
+            # RA & Dec
+            coord = SkyCoord(ra=float(entry['ra']) * u.deg, dec=float(entry['dec']) * u.deg,
+                             frame='icrs')
+            wcs.fill_in_parameters(metadata, coord, boresight=args.boresight, roll_ref=float(entry['roll_angle']))
+
+            # Observational metadata
+            metadata['instrument']['optical_element'] = entry['optical_element']
+            metadata['exposure']['ma_table_number'] = int(entry['ma_table_number'])
+            ma_table = parameters.ma_table[ metadata['exposure']['ma_table_number']]
```
schlafly, line 150: For this kind of stuff, I think you're right that we should pull out common code in romanisim-make-image and here to avoid duplication.  I think that either means making a new file in the module that has some of these routines and romanisim-make-image a thinner wrapper, or giving romanisim-make-image a --obslist argument that triggers a loop like this one.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                        spcript_file.write(line+'\n')
+
+                    else:
+                        # Obtain WCS catalog
+                        if args.catalog is None:
+                            twcs = wcs.get_wcs(metadata, usecrds=args.usecrds)
+                            rd_sca = twcs.toWorld(galsim.PositionD(
+                                galsim.roman.n_pix / 2, galsim.roman.n_pix / 2))
+                            cat = catalog.make_dummy_table_catalog(
+                                rd_sca, bandpasses=[entry['optical_element']], nobj=args.nobj)
```
schlafly, line 210: We don't want to make a new catalog for each entry / SCA.  That would mean that two images seeing the same sky would simulate different catalogs.  Probably the easiest thing to do is to just not support catalog == None.  Alternatively we need to generate a fake catalog outside both of these loops that covers the full footprint being simulated.  That seems a little messy, but we could imagine doing it.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+
+                # If SCA value is within 1-18, only run once
+                # otherwise loop over all detectors
+                while (sca <= NUMBER_OF_DETECTORS):
+                    # Detector
+                    metadata['instrument']['detector'] = f'WFI{sca:02d}'
+
+                    # Create output file name
+                    output_file_name = f"roman_simulated_{entry['ra']}_{entry['dec']}_{date}_{entry['optical_element']}_{entry_idx + 1}_{sca}.asdf"
+
```
schlafly, line 172: Let's leave off the ra/dec/date and just do entry_idx, filter, and sca.  But let's let the user configure that prefix (roman_simulated).  We should also probably include the "suffix" in the name; uncal for l1 and cal for l2.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,267 @@
...
+                            cat = Table.read(args.catalog)
+
+                        # If selected, apply persistence
+                        if args.previous:
+                            # Use provided Persistence file for first SCA (for each entry) only
+                            # after that, use previously created file
+                            if args.sca > 0 and sca > 1:
+                                persist = persistence.Persistence.read(previous_file_name)
+                            else:
+                                persist = persistence.Persistence.read(args.previous)
```
schlafly, line 224: Persistence is specific to each detector.  We need to pass in here the previous image simulated in the appropriate SCA.  The easiest approach is probably to make previous_file_name a dict, to fill in previous_file_name[sca] = output_file_name, and to do
prevfn = previous_file_name.get(sca, None)
persist = persist.Persistence() if prevfn is None else Persistence.read(prevfn)

schlafly (commented):
Sorry, I marked this up a lot!  A lot of little things, but I think the most important one is the one you brought up on slack, that we should make sure the common code in romanisim-make-image and romanisim-make-stack are in fact common code rather than duplicated.  I think my default approach would probably be to add an argument to romanisim-make-image to accept a --obslist as input that tells it where to grab all of the other inputs (and then to complain if incompatible arguments are set).  But a similarly good alternative would be to factor out the common code in romanisim-make-image and romanisim-make-stack so that they don't duplicate.

scripts/romanisim-make-stack
PaulHuwe: Ah, I thought that there initial latent image would be something else. We should discuss this.

scripts/romanisim-make-stack
PaulHuwe: I changed this in order to be able to use singe letter switches "-s" vs "-r". If you disagree with this, we can change it, but I like having the option of simple script options if possible. (I would actually rather change romanisim-make-image version to match.)

scripts/romanisim-make-stack
PaulHuwe: It does default to None (technically False) by not being activated - the reason it is what it is is in order to have '-m'.

scripts/romanisim-make-stack
PaulHuwe: It should be, as the offset is reset for each new entry (and then added to per exposure).

scripts/romanisim-make-stack
PaulHuwe: We could just require that any specified catalog spans the entire filed surveyed by the input pointings?

scripts/romanisim-make-stack
PaulHuwe: Agreed, I shall make it so.

scripts/romanisim-make-stack
PaulHuwe: Okay and agreed - the latter is on my TODO list.

scripts/romanisim-make-stack
schlafly: I really just want them to match.  Another option would be --seed and -S to avoid changing the interface, but your call.

scripts/romanisim-make-stack
schlafly: My point is really to allow this argument to do double duty as the name of the script to make (if one wants a script), rather than be true / false.

scripts/romanisim-make-stack
PaulHuwe: Yeah, it was on my TODO list during cleanup. Moved.

scripts/romanisim-make-stack
PaulHuwe: entry_idx should be exp_idx (exposure index); the time offset grows with each exposure and then gets reset once we go to the next entry / line.

PaulHuwe: "In the middle loop supposed to be, e.g., taking 10 exposures at the same place?"

This is exactly it. I figured that was what we were going to be given, otherwise, we wouldn't need overhead time (since if they listed one per line for every exposure, the time would match the start time of each). FYI, here is the input CSV I have been using:

```
ra, dec, roll_angle, optical_element, date, exposures, overhead_time, ma_table_number
270.0, 66.0, 0, F087, 2026-01-01T00:00:00, 4, 90, 1
270.0, 66.0, 45, F213, 2026-01-02T00:00:00, 2, 300, 1
273.0, 62.0, 10, F184, 2026-02-01T00:00:00, 3, 90, 1
273.0, 62.0, 55, F158, 2026-02-04T00:00:00, 1, 120, 1
270.0, 66.0, 0, F146, 2026-06-01T00:00:00, 2, 90, 1
270.0, 66.0, 60, F062, 2026-06-01T12:00:00, 4, 180, 1
273.0, 66.0, 10, F129, 2026-06-02T00:00:00, 3, 90, 1
273.0, 66.0, 55, F106, 2026-06-02T08:00:00, 1, 120, 1
```

scripts/romanisim-make-stack
PaulHuwe: I have changed this to a binary choice and removed persistence for the first image, as requested.

scripts/romanisim-make-stack
schlafly: Thanks, okay.

Let's aim for this format:
https://jira.stsci.edu/browse/APT-93485
In the context of that format, the 'duration' is the number of seconds that the exposure takes plus any overheads. I want to avoid our thinking about things like time needed to change the filter wheel, or how fast Roman can slew or spin or whatever, or how long settling + guide star acquisition + ... takes.

So we'll take one --date on the command line and add the duration to it each time to get the date of the next exposure.

You're right that an alternative would be to just do a separate date on each row.  Let's follow what I told them, though.  Let's not do repeated exposures at each location in either case.

PaulHuwe: Do you want date in the integer format of romanisim-make-image, UTC string, or something else? I am obviously a fan of the latter, but I will defer to your preference.

scripts/romanisim-make-stack
PaulHuwe: Implemented.

scripts/romanisim-make-stack
PaulHuwe: Implemented.

scripts/romanisim-make-stack
PaulHuwe: I moved this block to the beginning, and it centers on the midpoints of the RA & DEC range in the input file, with a radius spanning the range + FOV.

scripts/romanisim-make-stack
PaulHuwe: Refactored to use Tables.

scripts/romanisim-make-stack
PaulHuwe: Made them the same.

scripts/romanisim-make-stack
PaulHuwe: Implemented.

scripts/romanisim-make-stack
PaulHuwe: Adjustment mentioned above.

scripts/romanisim-make-stack
PaulHuwe: All three implemented.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,302 @@
...
+from astropy.time import Time
+import galsim
+from romanisim import wcs, persistence
+from romanisim.parameters import default_parameters_dictionary
+from romanisim import log
+from romanisim import parameters
+import ris_make_utils as ris
+
+# Radial distance from WFI_CEN to extend object generation in order to cover FOV (degrees)
+WFS_FOV = 0.6
```
PaulHuwe, line 17: I need to move this to parameters.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,302 @@
...
+    ra_diff = max_ra - min_ra
+    dec_diff = max_dec - min_dec
+
+    if args.debug:
+        print(f"max_ra = {max_ra}")
+        print(f"min_ra = {min_ra}")
+        print(f"ra_diff = {ra_diff}")
+        print(f"max_dec = {max_dec}")
+        print(f"min_dec = {min_dec}")
+        print(f"dec_diff = {dec_diff}")
```
schlafly, line 149: I don't use it a lot, but we have a log object, and these could be log.debug(...) without a special new debug argument, right?

I worry a bit about these ra/dec box calculations.  If someone wants to observe two images near the celestial pole, this ends up generating a catalog for the full sky?

I think to my sensibilities I just would not support "we'll make up a simple catalog for you" mode for this script.  The present default will do 1k objects over the full "survey area".  The current 1k / SCA is already pretty anemic, but 1k per field of view or larger will be very few objects.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,302 @@
...
+                        type=str,
+                        metavar='roman_simulated',
+                        default='roman_simulated',
+                        help='Filename prefix for simulated files.')
+
+    # Persistence
+    parser.add_argument('-p', '--previous',
+                        action='store_true',
+                        help=(
+                            'Enable persistence modeling of previous files for exposures after the first in each SCA.'))
```
schlafly, line 91: Probably rename persistence?

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,302 @@
...
+    # Date & Time argument
+    parser.add_argument('-d', '--date',
+                        type=str,
+                        default='2026-01-01T00:00:00',
+                        help=('UTC Date and Time of observation to simulate in ISOT format.'))
+
+    # Flux factor argument
+    parser.add_argument('-f', '--flux_factor',
+                        type=float,
+                        help='An optional flux factor for generated objects to make them brighter.')
```
schlafly, line 57: The way to do this is to control the magnitude limit in the catalog generation.  Let's not put a separate mechanism here.  I could imagine plumbing more into the catalog generation call, but I think we probably want to go the opposite direction and force people to generate a real catalog to use with their simulations rather than rely on an automatic made up one.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,302 @@
...
+                             usecrds=args.usecrds, x=(ra_center * galsim.degrees),
+                             y=(dec_center * galsim.degrees), radius=radius)
+
+    if args.debug:
+        print(f"cat_max_ra = {max(cat['ra'])}")
+        print(f"cat_min_ra = {min(cat['ra'])}")
+        print(f"cat_ra_diff = {max(cat['ra']) - min(cat['ra'])}")
+        print(f"cat_max_dec = {max(cat['dec'])}")
+        print(f"cat_min_dec = {min(cat['dec'])}")
+        print(f"cat_dec_diff = {max(cat['dec']) - min(cat['dec'])}")
```
schlafly, line 172: Do we think this is all still useful?

schlafly (approved):
This looks good.  I made a few additional minor comments in line.

I think the biggest one is that I think we should probably get rid of "we'll generate a catalog for you" input, or we need to continue making that better.  I don't think the current logic will work well around the celestial pole, and I worry that we should build in some more smarts about the number of objects (current number is probably way too low, and the number should likely scale with the area, etc.).

I also don't use the log.debug(...) interface a lot, but we did set up a romanisim.log object and you are doing a lot of debug work, so it's probably better to use that interface than to introduce a new --debug argument and print statements?

scripts/romanisim-make-stack
PaulHuwe: Done.

scripts/romanisim-make-stack
PaulHuwe: Sounds good - removing.

scripts/romanisim-make-stack
```diff
@@ -0,0 +1,302 @@
...
+
+    # Adjust brightness of generated objects before writing catalog
+    if args.flux_factor:
+        cat[cat.keys()[-1]] *= args.flux_factor
```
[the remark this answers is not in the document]
PaulHuwe: Removed with flux factor option removal.

scripts/romanisim-make-stack
PaulHuwe: You are right about the RA/DEC, though fixing that should be too difficult. Regarding the default objects, yeah, I was thinking that should be set to an object density, and then multiplied by the area to get an object total. If you like this idea, I can make a new ticket / issue to tackle them and remove them from this PR.

I will shift the debug stuff to logging.

scripts/romanisim-make-stack
PaulHuwe: Removed.

scripts/romanisim-make-stack
schlafly: Yeah, I think that would be the right way to go.  But more broadly, I think we want to make people specify a scene / catalog.  I think it's kind of nice having a "no scene" feature for romanisim-make-image just as the simplest possible entry point for people, but I don't think we want to try to make that feature more broadly.  I feel like if you've gotten to the point where you're simulating a full observational program it makes sense to make you specify a real catalog.

So I'd advocate just for removing that code and feature from the "stack" mode rather than trying to build out everything we'd want.

PaulHuwe: Okay, I will remove nobj, make a catalog file required, and gut that stuff out of the code.

PR #63, opened 2023-07-05
Ensure no jitter is simulated by WebbPSF

bmorris3:
WebbPSF simulates telescope jitter by default ([see docs](https://webbpsf.readthedocs.io/en/latest/usage.html#simulating-telescope-jitter)):
```python
>>> import webbpsf
>>> wfi = webbpsf.WFI()
>>> wfi.options
{'jitter': 'gaussian', 'jitter_sigma': 0.012}
```
...which may or may not be what `romanisim` users want. This PR ensures there's no jitter simulated by default. Alternatively, we could add these as kwargs for `make_psf` instead.

### DIFF
--- a/romanisim/psf.py
+++ b/romanisim/psf.py
@@ -77,6 +77,9 @@ def make_psf(sca, filter_name, wcs=None, webbpsf=True, pix=None,
     wfi.filter = filter_name
     wfi.detector_position = pix
     oversample = kw.get('oversample', 4)
+    wfi.options['jitter'] = None
+    wfi.options['jitter_sigma'] = 0
+
     # webbpsf doesn't do distortion
     psf = wfi.calc_psf(oversample=oversample)
     gimg = galsim.Image(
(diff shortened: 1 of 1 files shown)


schlafly:
My tendency here is to keep the WebbPSF defaults; jitter is a real component of the PSF and if WebbPSF has a default number for it I wouldn't override it unless we have a better number.

bmorris3:
Fair. I'll update this PR by adding kwargs present with the WebbPSF defaults.

Worth keeping in mind – the default standard deviation of Gaussian jitter is 0.012 arcsec, which corresponds to 0.11 pixels.

PaulHuwe (approved):
LGTM, but I defer to @schlafly for final word.

schlafly:
What do you think about an alternative implementation where you do:
`wfi.options.update(**kw)`
and update the docs to say that additional arguments are passed to wfi.options or roman.getPSF?  My preference there is driven by wanting to follow WebbPSF if WebbPSF decides to change the default jitter model or amplitude, without needing to separately update in the simulator.  i.e., try to keep WebbPSF in charge of the PSF modeling as much as possible, but let the API handle more options there?

PR #64, opened 2023-07-05
Refactor ramp fitting code to use less memory.

schlafly:
This PR refactors the ramp fitting code to use less memory.  The underlying calculations are identical and the same results are obtained as before.  But the new version loops over the array pixel by pixel, doing the complete calculation for each pixel before moving onto the next, while the old version computed intermediate quantities for each pixel before combining them to get the complete results for the array.

The new version is slightly slower; doing a million pixels with 6 resultants takes something like 400 ms with this PR instead of 300 ms it the original version.  I suspect this is because of the less efficient memory access pattern.  Because the arrays are sequential spatially rather than temporally the new pattern which loops over each pixel is not very friendly for caching and read-ahead.  Maybe there's a world where we transpose the arrays first.  But I don't see a real need to do that to save a few seconds.

### DIFF
--- a/romanisim/ramp.py
+++ b/romanisim/ramp.py
@@ -549,7 +549,8 @@ def fit_ramps_casertano(resultants, dq, read_noise, ma_table):
     var.reshape(npix, 3, 2, 2)[..., 1, 1, 1] = (
         totval / (totweight ** 2 + (totweight == 0)))
 
-    var[..., 1, 1, 1] *= par[..., 1]  # multiply Poisson term by flux
+    # multiply Poisson term by flux.  Clip at zero; no negative Poisson variances.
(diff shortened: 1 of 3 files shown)


schlafly:
@stscieisenhamer , care to take a look at this?  I've been looking at memory consumption in romancal over the last week and I think the new ramp fitting & reference pixel correction codes become the tall poles following some modest changes to eliminate copies of datamodels.  This PR changes the simulator code to avoid making any large intermediate data structures (the former weight / resultant coefficient arrays).  We should do the same for the real stcal code.

stscieisenhamer (commented):
LGTM. Anything the reduces/centralizes code and arrays is a good.

PR #67, opened 2023-07-18
Misc deprecation fixes and test failures.

schlafly:
Fix some cosmetic issues and upcoming deprecations.
* types to f4 rather than f8 in catalogs where appropriate
* np.product -> np.prod
* raise an error if you try to get a persistence flux from times _before_ the persistence event occurred
* metadata['exposure']['start_time'] needs to be an astropy Time object.

### DIFF
--- a/romanisim/__init__.py
+++ b/romanisim/__init__.py
@@ -15,8 +15,8 @@ from pkg_resources import get_distribution, DistributionNotFound
 import logging
 
 log = logging.getLogger(__name__)
-logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s',
-                    datefmt='%Y-%m-%d %H:%M:%S')
+# logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s',
+#                     datefmt='%Y-%m-%d %H:%M:%S')
 log.setLevel(logging.INFO)
 
 try:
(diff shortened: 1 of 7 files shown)


schlafly:
And @stscieisenhamer , sorry, this will affect the stcal implementation as well.  The cython pyx module has two minor changes:
- an @cython.cpow(True) decorator
- an if nres <= 1: return(0, 0, 0)
that heads off a special case early.  Apparently some cython update changed the behavior of 0**0 and the simulator implementation started failing tests, so we'd want this in the stcal implementation as well.

romanisim/__init__.py
```diff
@@ -15,8 +15,8 @@
 import logging
 
 log = logging.getLogger(__name__)
-logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s',
-                    datefmt='%Y-%m-%d %H:%M:%S')
+# logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s',
+#                     datefmt='%Y-%m-%d %H:%M:%S')
```
PaulHuwe, line 19: Should these be deleted?

PaulHuwe (approved):
LGTM with one suggestion

romanisim/__init__.py
schlafly: I like those but they are interacting with stcal's logging in a real way when --usecrds is on.  I should include some more fulsome documentation but let's leave that for the moment.  Thanks!

PR #68, opened 2023-07-28
Add or improve tests for various simulator requirements.

schlafly:
* DMS217: Parametric source distributions.
* DMS218: Distortion & WCS.
* DMS229: Ramp from counts.
* DMS230: Poisson noise.

A few unrelated changes:
- initialize noise array with zeros instead of empty because galsim wants to cast it before filling it, and can try to cast invalid numbers and emit a warning.
- fix the units on some test_simulate(...) tests to avoid generating sources with 10^20 photons.

### DIFF
--- a/romanisim/l1.py
+++ b/romanisim/l1.py
@@ -377,7 +377,7 @@ def add_read_noise_to_resultants(resultants, tij, read_noise=None, rng=None,
     else:
         rng = galsim.GaussianDeviate(rng)
 
-    noise = np.empty(resultants.shape, dtype='f4')
+    noise = np.zeros(resultants.shape, dtype='f4')
     rng.generate(noise)
     if read_noise is None:
         read_noise = parameters.read_noise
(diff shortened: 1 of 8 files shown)


PaulHuwe (approved):
LGTM

schlafly:
@PaulHuwe , sorry, I need to return this to a draft.  When I went to do my usual run of romanisim-make-image against a bunch of different options, I got failures because when run without an explicit catalog, ris_make_utils.create_catalog was being called without a metadata argument and was causing problems.
https://github.com/spacetelescope/romanisim/pull/68/files#diff-62d8222aafdb5766d61cf5d8d6e68500f89cf42c95eb088bb07dedca170b6f56L63
It's a good reminder that I need to do some better end-to-end tests there.  I've now moved ris_make_utils into the main romanisim directory and added some tests.  I don't have any end-to-end tests of the stack module so you might need to make sure I haven't broken anything there.  But I'll look at this again Monday.

schlafly:
@PaulHuwe , can you give this another look?  I added some ris_make_utils tests and changed some of the call signatures a bit to make them easier to test.  Thanks!

PR #70, opened 2023-08-15
RCAL-582 Nonlinearity CRDS Support in Simulator

PaulHuwe:
Replaced the QE-like statistics presently used for inverse nonlinearity application with files pulled from CRDS. Added tests to confirm that this application.


romanisim/l1.py
```diff
@@ -282,48 +288,72 @@ def apportion_counts_to_resultants(
...
             resultant_counts += counts_so_far + instrumental_so_far
+            # if linearity is not None:
+            #     ki_denominator += efficiency * pij_per_read[i][j]
+            # readnum += 1
+
+            # Apply linearity
             if linearity is not None:
-                ki_denominator += efficiency * pij_per_read[i][j]
-            readnum += 1
+                resultant_counts += linearity.apply(counts_so_far + instrumental_so_far, electrons=True)
```
schlafly, line 354: This is applying the counts twice if linearity is not None.  This needs to be more like:
if not None:
    resultant_counts += linearity.apply(...)
else:
    resultant_counts += ...

romanisim/image.py
```diff
@@ -696,8 +701,7 @@ def simulate(metadata, objlist,
     else:
         l1, l1dq = romanisim.l1.make_l1(
             counts, ma_table_number, read_noise=read_noise, rng=rng, gain=gain,
-            crparam=crparam,
-            linearity=linearity, tstart=image_mod.meta.exposure.start_time,
+            crparam=crparam, linearity=inverse_linearity, tstart=image_mod.meta.exposure.start_time,
```
schlafly, line 704: Let's change the name here; linearity = inverse_linearity is asking for trouble..

romanisim/nonlinearity.py
```diff
@@ -74,76 +78,20 @@ def repair_coefficients(coeffs):
         linear values with slopes of unity.
     """
     res = coeffs.copy()
+
+    if inverse:
```
schlafly, line 82: I think you're right that we can just get rid of inverse heal and elsewhere.  I think I was imagining the NL object having both the inverse and forward coefficients in it, so there would be one linearity object that knew how to reverse itself.  But your design is also good.

romanisim/nonlinearity.py
```diff
@@ -199,35 +148,19 @@ class NL:
...
-
-        Returns
-        -------
-        efficiency : np.ndarray[nx, ny] (float)
-            The efficiency of each pixel at the current count level.
-        """
-        return efficiency(counts, self.derivcoeffs, reversed=True)
-
-    def correct(self, counts):
+    def apply(self, counts, electrons=False, reversed=False):
```
schlafly, line 163: electrons needs documentation.  It's what you should use if 'counts' is actually in electrons but the NL coefficients are designed for DN, applying and unapplying the gain.

romanisim/tests/test_l1.py
```diff
@@ -84,6 +84,39 @@ def test_apportion_counts_to_resultants():
...
+    coeffs = np.asfarray([1.0, 0.7, 3.0e-6, 5.0e-12])
+    lin_coeffs = np.tile(coeffs[:,np.newaxis,np.newaxis], (1, 100, 100))
+    lin_coeffs[:, 0:50, :] *= 2.0
+    gain = 4.0
+    counts[0,0] = counts[0,99] = counts[99,0] = counts[99,99] = 11.0
+
+    linearity = nonlinearity.NL(lin_coeffs, gain)
+
+    for tij in tijlist:
+        resultants, dq = l1.apportion_counts_to_resultants(counts.copy(), tij.copy())
```
schlafly, line 100: Do counts or tij get touched here?  Are these copys necessary?

romanisim/nonlinearity.py
```diff
@@ -199,35 +148,19 @@ class NL:
...
         """
+        self.inverse = inverse
         self.coeffs = repair_coefficients(coeffs)
-        self.derivcoeffs = derivative(self.coeffs)
-        self.coeffs = self.coeffs[::-1, ...].copy()
-        self.derivcoeffs = self.derivcoeffs[::-1, ...].copy()
-
-    def efficiency(self, counts):
-        """Compute the efficiency of photon counting given a count level.
+        self.gain = gain
```
schlafly, line 161: It looks like I used to do self.coeffs = self.coeffs[::-1, ...].copy() around here; was that dropped intentionally?

romanisim/tests/test_l1.py
```diff
@@ -84,6 +84,39 @@ def test_apportion_counts_to_resultants():
...
+        assert np.all(res2 >= 0)
+
+        assert np.all(res2 != resultants)
+
+        assert np.all(res2[-1:1, 0, 0] != res2[-1:1, 0, 99])
+        assert np.all(res2[:, 0, 0] != res2[:, 99, 0])
+        assert np.all(res2[:, 0, 0] != res2[:, 99, 99])
+        assert np.all(res2[:, 0, 99] != res2[:, 99, 0])
+        assert np.all(res2[:, 0, 99] != res2[:, 99, 99])
+        assert np.all(res2[:-1, 99, 0] != res2[:-1, 99, 99])
```
schlafly, line 116: It's not true that any of these need to be true for any possible set of random numbers or tij?  e.g. for the second one, res2[:, 0, 0] compares the resultants for one corner with the ones for another.  You gave them the same total number of counts.  You gave them pretty different linearity coefficients, but if the first resultant in the one corner happens to get more / less photons early on, that could lead to a match?

It looks to me like I'm expecting the first coefficient in the file to be around 0 and the second to be around 1, and the higher coefficients to be small.  Your example coefficients don't do that.  Is that intentional?

Some tests that you could do that would make more sense to me:
- generate res, res2 with the same random number generator.
- give res2 some simple linearity coefficients, e.g., [0, 0.5, 0, ...].  This effectively changes the gain by a factor of 2.
- check that np.all(res2 <= res) and np.any(res2 < res), which should be ~guaranteed (at least one resultant got fewer counts (since a count now corresponds to two electrons), and all of them are less than or equal.

I agree with you that we also want a simple test that the coefficients and their inverse invert.  e.g., coeffs = [0, 2, 0, 0].  coeffs2 = [0, 0.5, 0, 0].  nl1.apply(nl2.apply(values)) == values.

romanisim/tests/test_linear.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@pytest.mark.parametrize("inverse", INVERSE)
+def test_repair_coeffs(inverse):
+    counts = np.random.poisson(100, size=(100, 100))
+    coeffs = np.asfarray([1.0, 0.7, 3.0e-6, 5.0e-12])
+    lin_coeffs = np.tile(coeffs[:, np.newaxis, np.newaxis], (1, 100, 100))
+    lin_coeffs[:, 0:50, :] *= 2.0
+    lin_coeffs[:, 1, 1] *= 0
+    lin_coeffs[2, 22, 22] = np.nan
```
schlafly, line 65: Let's add a comment saying that these are invalid coefficients and so these pixels should get no correction.

romanisim/tests/test_linear.py
```diff
@@ -0,0 +1,185 @@
...
+    gain = 4.0 * u.electron / u.DN
+
+    linearity = nonlinearity.NL(lin_coeffs, gain, inverse)
+
+    res = linearity.apply(counts)
+
+    assert res[1, 1] == counts[1, 1]
+    assert res[22, 22] == counts[22, 22]
+    assert np.all(res[23:,23:] != counts[23:,23:])
```
schlafly, line 74: This is really "all of the array except the two pixels I messed with"?  Maybe np.sum(res != counts) == np.prod(counts.shape) - 2 plus a comment?

romanisim/tests/test_linear.py
```diff
@@ -0,0 +1,185 @@
...
+        reffiles['inverselinearity'])
+    linearity_model = roman_datamodels.datamodels.LinearityRefModel(
+        reffiles['linearity'])
+
+    gain = 1.0
+
+    inverse_linearity = nonlinearity.NL(
+        inverse_linearity_model.coeffs[:,0:4088,0:4088], gain=gain)
+    linearity = nonlinearity.NL(
+        linearity_model.coeffs[:,0:4088,0:4088], gain=gain)
```
schlafly, line 167: This should probably be coeffs[:,4:,:-4], etc.

romanisim/tests/test_linear.py
```diff
@@ -0,0 +1,185 @@
...
+        metadata, cat, usecrds=True,
+        webbpsf=True, level=0,
+        rng=rng)
+
+    level_0_inv = inverse_linearity.apply(level_0['data'])
+
+    level_0_lin = linearity.apply(level_0_inv)
+
+    # Divide initial counts before and after inverse & NL application
+    div = level_0['data'][np.isfinite(level_0['data'])] / level_0_lin[np.isfinite(level_0_lin)]
```
schlafly, line 181: This is a good test, but there's no reason to use a real image for it?  i.e., something like:
```
counts = np.poisson(...)
nl_counts = inverse_linearity.apply(counts)
assert np.all(counts == lineary.apply(nl_counts))
```
Then you can get rid of all of the bits of the test that go into making that level 0 file (catalogs, metadata, ...).

schlafly (changes requested):
This looks good.  I left a lot of comments in line.  Some more comments:
- Can you update the module level comments in the nonlinearity module?  There's a current long description that is completely out of date now and needs to be replaced with your approach.
- Your tests look more like other tests than mine do.  But the degree of code duplication hurts me a little.  e.g., I don't think the current set of coeffs make sense (should start with 0, I think?), and now we're going to be updating that everywhere.

romanisim/l1.py
```diff
@@ -256,17 +257,22 @@ def apportion_counts_to_resultants(
...
-    efficiency = 1
-    if linearity is not None:
-        pij_per_read = tij_to_pij(tij)
-        ki_numerator = np.cumsum(pij_per_read)
-        ki_denominator = np.zeros(counts_so_far.shape, dtype='f4')
+    # efficiency = 1
+    # if linearity is not None:
+    #     pij_per_read = tij_to_pij(tij)
+    #     ki_numerator = np.cumsum(pij_per_read)
+    #     ki_denominator = np.zeros(counts_so_far.shape, dtype='f4')
```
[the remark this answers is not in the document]
PaulHuwe: Deleted.

romanisim/l1.py
PaulHuwe: Oops.. line moved.

romanisim/nonlinearity.py
PaulHuwe: Yes, just to keep the `reverse` keywords defaulting to false. I can swap the logic around, if you prefer.

romanisim/tests/test_l1.py
PaulHuwe: Code left in from debugging. `copy`s removed.

romanisim/tests/test_linear.py
PaulHuwe: Changed.

romanisim/tests/test_linear.py
PaulHuwe: Changed.

romanisim/image.py
PaulHuwe: I changed it here and throughout l1 for clarity.

romanisim/nonlinearity.py
PaulHuwe: removed.

romanisim/tests/test_linear.py
PaulHuwe: Added.

romanisim/tests/test_linear.py
PaulHuwe: Trimmed down.

romanisim/nonlinearity.py
PaulHuwe: Added.

schlafly:
It looks like tests are sometimes failing, maybe due to a tolerance thing?

romanisim/nonlinearity.py
```diff
@@ -187,57 +129,55 @@ def correct(counts, coeffs, reversed=False):
...
-        return efficiency(counts, self.derivcoeffs, reversed=True)
-
-    def correct(self, counts):
+        # self.coeffs = repair_coefficients(coeffs)
+        self.gain = gain
+        if dq is not None:
+            self.coeffs, self.dq = repair_coefficients(coeffs, dq)
+        else:
+            self.coeffs = repair_coefficients(coeffs)
+            self.dq = None
```
schlafly, line 155: Is self.dq ever used?  I guess ideally in make_l1 or something the problematic linearity bits should get or'ed into the image dq bits?

romanisim/nonlinearity.py
schlafly: Okay, I see now.  Some part of me felt like it was better to "correct" the mis-ordering on read-in and then have apply the coefficients in the "fast" way (i.e., without reversing again.  I doubt that's actually faster, though.  Okay, sounds good, no action needed.

romanisim/tests/test_l1.py
```diff
@@ -84,6 +84,33 @@ def test_apportion_counts_to_resultants():
...
+    for tij in tijlist:
+        resultants, dq = l1.apportion_counts_to_resultants(counts.copy(), tij.copy(), rng=rng)
+
+        res2, dq2 = l1.apportion_counts_to_resultants(counts.copy(), tij.copy(), rng=rng,
+                                                      inv_linearity=inv_linearity)
+
+        # Ensure that the linear coefficients were actually applied
+        assert np.all(res2 >= 0)
+        assert np.any(res2.sum(axis=0) < resultants.sum(axis=0))
+        assert np.all(res2.sum(axis=0) <= resultants.sum(axis=0))
```
schlafly, line 109: Can we require something like the median of the difference between the first and last resultant changes by something roughly like a factor of two?  Might need to make this more like 100x100 than 4x4 for that to come out decently closely.

romanisim/tests/test_linear.py
```diff
@@ -0,0 +1,144 @@
...
+ma_table_list = [
+    [[1, 10], [11, 1], [12, 10], [30, 1], [40, 5], [50, 100]],
+    [[1, 1]],
+    [[1, 1], [10, 1]],
+]
+
+
+def test_linear_apply():
+    counts = np.random.poisson(100, size=(100, 100))
+    coeffs = np.asfarray([1.0, 0.7, 3.0e-6, 5.0e-12])
```
schlafly, line 33: I feel like the first coeff should almost always be close to zero.  Is that not the case in the real linearity files?  This is the ~pedestal / y-intercept, right?  I mean, it's fine to make it whatever, but it looks weird with a pedestal.

romanisim/tests/test_linear.py
```diff
@@ -0,0 +1,144 @@
...
+        linearity_model.coeffs[:,4:-4,4:-4], gain=1.0)
+
+    counts = np.random.poisson(16000, size=(4088, 4088))
+    level_0_inv = inverse_linearity.apply(counts)
+    level_0_lin = linearity.apply(level_0_inv)
+
+    # Divide initial counts before and after inverse & NL application
+    div = (counts[np.isfinite(counts)] / level_0_lin[np.isfinite(level_0_lin)]).flatten()
+
+    assert np.isclose(np.average(div), 1.0, atol=0.1)
```
schlafly, line 144: It doesn't look like these were very close?  I think the inverse linearity is literally intended to be the inverse of the linearity, so I was hoping for pretty good agreement.  What atol / rtol is needed if you wanted the test to be
np.isclose(counts[m], level_0_lin[m])?

schlafly (changes requested):
Hi Paul,

Few more minor comments, sorry.  This is looking good though.

romanisim/l1.py
```diff
@@ -256,74 +257,63 @@ def apportion_counts_to_resultants(
...
-    # The denominator depends on the realized efficiency of each pixel
-    # and so must be built up during the read.
-
-    readnum = 0
-    nlflag = np.zeros(counts.shape, dtype='bool')
-
+    # Set initial instrument counts
     instrumental_so_far = 0
+
+    # Maybe this can be un-iffed and set above?
```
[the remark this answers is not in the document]
PaulHuwe: Okay - comment removed.

romanisim/nonlinearity.py
PaulHuwe: Refactored this a bit.

romanisim/tests/test_l1.py
PaulHuwe: Done, and it matches well.

romanisim/tests/test_linear.py
PaulHuwe: I stole these numbers from STCAL, so it is possible they were made with JWST in mind? I didn't care too strongly about the numbers themselves, just the the orders of magnitude and that each number was different. I can change these to something else, if you want?

romanisim/tests/test_linear.py
schlafly: I just pushed an update here that looks pretty good to me.  For the ~50% of pixels that aren't obviously bonkers, the corrections look pretty good, and invert at the 1e-6 level at your chosen flux level.  I've put in a fractional uncertainty rms of 1e-5.

romanisim/tests/test_linear.py
schlafly: It just seems to me like we should pick numbers that "look like" linearity coefficients.  Here's an image of the log10 of the absolute value some of the linearity coefficients
image
The first coefficient is always zero and gets NaNed to white here.  The second coefficient is always very close to one.  Higher coefficients drop off very quickly (10^-5, 10^-10, 10^-15, ...).  It's of course okay to choose any set of coefficients, but when I read them I get the impression that you think that the first coefficient corresponds to the linear term rather than the constant term (since it is the linear terms in the reference file that are close to 1).  It does look to me like the stcal code follows the same convention (just a polynomial, with coefficients starting from a constant going to x^(ncoeff - 1)), so I'm surprised that the stcal tests would use these numbers.  But I don't think we need to copy them.

romanisim/tests/test_linear.py
PaulHuwe: Okay, I have downloaded it and merged it with my local work.

romanisim/tests/test_linear.py
PaulHuwe: Changed.

schlafly (approved):
This looks good to me.  Feel free to merge when you're ready.  The one somewhat important thing that I touched was to include the gain in the NL(....) invocation in image.simulate.  Otherwise we aren't including it when doing the NL application in apportioning photons to reads and that's bad.

PR #71, opened 2023-08-16
Update for webbpsf 1.2.1.

schlafly:
This updates the webbpsf data files used in CI to the latest versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
           fetch-depth: 0
       - id: data
         run: |
-          echo "webbpsf_url=https://stsci.box.com/shared/static/n1fealx9q0m6sdnass6wnyfikvxtc0zz.gz" >> $GITHUB_OUTPUT
+          echo "webbpsf_url=https://stsci.box.com/shared/static/0ojjfg3cieqdpd18vl1bjnpe63r82dx8.gz" >> $GITHUB_OUTPUT
           echo "galsim_url=https://github.com/GalSim-developers/GalSim/raw/releases/2.4/examples/data/" >> $GITHUB_OUTPUT
           echo "path=/tmp/data" >> $GITHUB_OUTPUT
       - run: |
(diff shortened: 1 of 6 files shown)


schlafly:
@zacharyburnett , would you mind taking a look at this?  webbpsf 1.2.1 came out and needed a corresponding bump to the data file that the CI downloads.  I tried to make that change there but somehow the CI is still seeing 1.1.1 (error message from CI: OSError: WebbPSF data package has version 1.1.1, but 1.2.1 is needed).  I can look at the CI logs and see that it is downloading the new files in the CI for this PR, and when I download those files myself they work fine with webbpsf 1.2.1.  Do I need to do something special to "clear the cache" or something to get it to see the right files, or have I screwed something else up?  Thanks!

schlafly:
After clearing the cache I got this to work better.  I suspect the hashing going on in the CI is not working quite right and it should have been recognizing that the new files were different but it wasn't actually doing that.  There is still some additional issue where the tox codecov runs are failing, but I don't understand that.  @zacharyburnett , sorry, I can't make heads or tails of this one.

schlafly:
Okay, this is now solved.  At least four relevant issues:
- The WebbPSF link needed to be updated due to their 1.2.1 release.  It seems to me that they could have a -latest-released link that could reduce some churn there?
- The hashing used to make the cache doesn't look quite right and didn't recognize that the new file downloaded as part of 1.2.1 wasn't identical to the old one (?).  The tox jobs then got the old data and were still out of date after updating the links and downloading the now files.  Clearing the cache in the actions tab caused this to grab a new version and succeed.
- sphinx had an update that was breaking the theme used for the readthedocs build.  I pinned to versions less than 7.
- tox had an update that identified an issue where we weren't appropriately listing all of the tox environments correctly.  This PR tweaks the names of the tox environments called by the ci to be consistent, and adds an environment to the tox.ini file.
I think that's everything!  But I don't really know what I'm doing with any of these.

PaulHuwe (approved):
LGTM - I defer to @zacharyburnett on the CI issues.

PR #72, opened 2023-08-25
Add test artifacts to B11 tests.

schlafly:
This PR adds some asdf binary files with the various images & test statistics computed during the B11 tests.


schlafly:
@PaulHuwe , care to review briefly?  This adds some ~artifacts to the regression tests for B11, basically writing out the things we already compute so that they're available for inspection by others.

PR #78, opened 2023-09-19
RCAL-580 Input Config File

PaulHuwe:
Adding support for an input configuration file that would override romanisim.parameters and allow for changing parameters like metadata, general default constants, and API options.

### DIFF
--- a/romanisim/catalog.py
+++ b/romanisim/catalog.py
@@ -10,7 +10,7 @@ from galsim import roman
 from astropy import coordinates
 from astropy import table
 from astropy import units as u
-from . import util, parameters
+from . import util
 import romanisim.bandpass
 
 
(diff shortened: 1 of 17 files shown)


PaulHuwe:
Sample configuration file: 
[params.yaml.txt](https://github.com/spacetelescope/romanisim/files/12664700/params.yaml.txt)

schlafly:
One dumb question: do you know if this works?  I was expecting to have to do
```
def function(x, k1=None, k2=None, ...):
    if k1 is None:
        k1 = parameters['k1']
```
I like what you've written more, except that I had thought that the function default values were set once at ~import time, so that if we later update parameters.values those changes don't get reflected in the default values.

Otherwise looks good.  I don't think it's necessary to make any changes to the defaults for the catalog routines; I think we want to encourage people to make real catalogs rather than relying on the canned distributions.

PaulHuwe:
I am happy to remove the catalog parameters - a nice addition of this setup is that it is very easy to control what is and isn't modifiable simply by what is in the parameter file vs. in the source files.

PaulHuwe:
As for testing, I feel that this would best be tested with regression tests, however I am open to other ideas.

schlafly (approved):
Looks good to me, thanks!

PR #79, opened 2023-09-27
RCAL-645 Override Reference Files

PaulHuwe:
Added the ability to provide reference file names instead of calling CRDS for matched reference files via the parameters file. Also allows for overriding with flat values, if not using reference files.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -635,7 +635,7 @@ def simulate(metadata, objlist,
         import crds
         refnames_lst = ['readnoise', 'dark', 'gain', 'inverselinearity', 'linearity', 'saturation']
 
-        for key, value in parameters.reference_file_names.items():
+        for key, value in parameters.reference_data.items():
             if (key in refnames_lst) and (value is not None):
                 reffiles[key] = value
                 refnames_lst.remove(key)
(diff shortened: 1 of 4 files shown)


PaulHuwe:
Tested with the attached parameter files:
[params_ref.yaml.txt](https://github.com/spacetelescope/romanisim/files/12733355/params_ref.yaml.txt)
[params_partial_ref.yaml.txt](https://github.com/spacetelescope/romanisim/files/12733356/params_partial_ref.yaml.txt)

romanisim/ris_make_utils.py
```diff
@@ -123,7 +124,13 @@ def create_catalog(metadata=None, catalog_name=None, bandpasses=['F087'],
...
         # Create a catalog from scratch
         # Create wcs object
-        twcs = wcs.get_wcs(metadata, usecrds=usecrds)
+        distortion_file = parameters.reference_file_names["distortion"]
+        if distortion_file is not None:
+            dist_model = roman_datamodels.datamodels.DistortionRefModel(distortion_file)
+            distortion = dist_model.coordinate_distortion_transform
+        else:
+            distortion = None
+        twcs = wcs.get_wcs(metadata, usecrds=usecrds, distortion=distortion)
```
schlafly, line 133: This is my mistake, but let's take this opportunity to document the distortion argument of get_wcs.

romanisim/parameters.py
```diff
@@ -26,6 +26,16 @@
...
+reference_file_names = {
+    "dark": None,
+    "distortion": None,
+    "flat": None,
+    "gain": None,
+    "inverselinearity": None,
+    "linearity": None,
+    "readnoise": None,
+    "saturation": None
+}
```
schlafly, line 38: Since we're allowing this to be either a file name or a constant, maybe reference_data?  If we moved some of the defaults in here, would that simplify the logic in image.py?  i.e.,
reference_data = { 'dark': galsim.roman.dark_current, 'flat': 1, gain: 1 u.electron / u.s, ... }

romanisim/image.py
```diff
@@ -684,14 +692,17 @@ def simulate(metadata, objlist,
...
-        darkrate = galsim.roman.dark_current
-        dark = None
-        gain = None
-        flat = 1
-        inv_linearity = None
-        linearity = None
-        saturation = None
+        read_noise = parameters.reference_file_names["readnoise"] \
+            if parameters.reference_file_names["readnoise"] is not None else galsim.roman.read_noise
+        darkrate = (parameters.reference_file_names["dark"] / exptime_tau) \
```
schlafly, line 697: For the case where this is a constant, let's just have the constant be the dark current rather than the last value in the dark reference file divided by tau.

schlafly (commented):
I like this implementation.  A few minor changes suggested below; let me know what you think.

romanisim/image.py
PaulHuwe: Changed to a new keyword.

romanisim/ris_make_utils.py
PaulHuwe: I noticed this as well, but forgot to address it. I have added an entry.

romanisim/parameters.py
PaulHuwe: I changed the name of the module as suggested. However, I am resistant to change the default values - I like using `None` as it doesn't "type hint," which a value would do.

schlafly (approved):
LGTM, thanks!

PR #80, opened 2023-09-29
Allow ramp fitting to use pixels marked with bad non-linearity.

schlafly:
This PR allows ramp fitting to proceed on pixels marked as having problematic linearity.

Formerly we had only two DQ bits---saturation and CR.  Ramp fitting ignored any pixel with a DQ bit so that we didn't include these in the ramps.  When we added the nonlinearity DQ bit, this made us stop fitting ramps on all of the pixels with problematic linearity determination, which is not what we want.

This adds a new parameters.dq_do_not_use set of flags that identifies the flags that should be used for determining which resultants to ignore in ramp fitting.

I also changed the name of the dq bit we're using from nonlinear to no_lin_corr, since I think the latter is a better match to what's happening in the simulator at the moment.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -125,10 +125,11 @@ def make_l2(resultants, ma_table, read_noise=None, gain=None, flat=None,
 
     if linearity is not None:
         # Update data quality array for linearty coefficients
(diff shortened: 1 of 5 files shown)


PaulHuwe (approved):
LGTM

PR #81, opened 2023-09-30
RCAL-628 SOC Test decorators

PaulHuwe:
Added SOC Test decorators for metrics logging via https://github.com/spacetelescope/metrics_logger


schlafly (approved):
Thanks Paul, looks good to me, merging.  Thanks for finding the cases where I had failed to put a comment in the docstring.

PR #83, opened 2023-10-04
Store v2/v3 in arcsec.

schlafly:
In the current code V2/V3 are populated in the metadata in degrees.  This changes that so that they are populated in arcsec, the usual units for V2/V3 coordinates, but they are converted to degrees before using them to construct the WCS.


PaulHuwe:
Maybe add a comment to parameters.py explaining the various units for the vN variables?

schlafly:
Thanks, I added a comment here.  I agree that there would be value in doing this more globally but I can't bring myself to look up the meaning of all of the e.g. persistence parameters.

Issue #88, opened 2023-10-30
read_pattern is not being set to match used ma table


stscieisenhamer:
The meta value `meta.exposure.read_pattern` is not being set to match the ma table in use. Example using the MA table "2", which, according to `parameters.py` is a 5-resultant table. However, `read_pattern` has six resultants. Use of any ma table results in the same `read_pattern`:

```
$ romanisim-make-image --level 1 --ma_table_number 2 random_uncal.asdf
$ python
>>> import roman_datamodels as rdm
>>> m = rdm.open('random_uncal.asdf')
>>> m.meta.exposure.read_pattern
LNode([[1], [2, 3], [4], [5, 6, 7, 8], [9, 10], [11]])
>>> m.data.shape
(5, 4096, 4096)
```

schlafly:
Oh yeah, we wrote this before read_pattern existed; read_pattern presumably always takes that garbage value.

PR #89, opened 2023-10-30
Add read_pattern metadata to simulated files.

schlafly:
This PR adds the read_pattern to simulated images.
Fixes https://github.com/spacetelescope/romanisim/issues/88.

### DIFF
--- a/romanisim/util.py
+++ b/romanisim/util.py
@@ -218,7 +218,7 @@ def add_more_metadata(metadata):
     metadata['exposure']['effective_exposure_time'] = openshuttertime
     metadata['exposure']['duration'] = openshuttertime
     read_pattern = l1.ma_table_to_tij(ma_table, read_time=1)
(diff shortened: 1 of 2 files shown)


romanisim/util.py
```diff
@@ -217,6 +217,8 @@ def add_more_metadata(metadata):
     metadata['exposure']['exposure_time'] = openshuttertime
     metadata['exposure']['effective_exposure_time'] = openshuttertime
     metadata['exposure']['duration'] = openshuttertime
+    read_pattern = l1.ma_table_to_tij(ma_table, read_time=1)
```
stscieisenhamer, line 220: `read_pattern` is being defined, but is not used? The next line re-calculates read_pattern and sets it directly in the meta.

schlafly: Thanks, this should be fixed now.

stscieisenhamer (approved):

PR #90, opened 2023-11-16
RCAL-663: Image Source Injection Test

PaulHuwe:
Added test for source injection into Level 2 Images.


romanisim/tests/test_image.py
```diff
@@ -635,3 +635,109 @@ def test_reference_file_crds_match(level):
...
+
+    # Create catalog of sources
+    twcs = wcs.get_wcs(metadata, usecrds=False)
+    rd_sca = twcs.toWorld(galsim.PositionD(
+        galsim.roman.n_pix / 2, galsim.roman.n_pix / 2))
+    cat = catalog.make_dummy_table_catalog(
+        rd_sca, bandpasses=[metadata['instrument']['optical_element']], nobj=10)
+
+    # Brighten sources
+    cat['F158'] = [flux * f158 for f158 in cat['F158']]
```
schlafly, line 669: Can we set the fluxes to specific values rather than brighten them?  It's possible for the fluxes to be large and a factor of 10k larger could be very large.  Maybe we should specify a seed when making the dummy catalog?

romanisim/tests/test_image.py
```diff
@@ -635,3 +635,109 @@ def test_reference_file_crds_match(level):
...
+
+@metrics_logger("DMS231")
+@pytest.mark.soctests
+def test_inject_source_into_image():
+    """Inject a source into an image.
+    Demonstrates DMS231.
+    """
+
+    # Set constants and metadata
+    galsim.roman.n_pix = 4088
```
schlafly, line 648: If we make this much smaller, does it cause problems?  usecrds=False, so it's probably okay?

romanisim/tests/test_image.py
```diff
@@ -635,3 +635,109 @@ def test_reference_file_crds_match(level):
...
+    cat = catalog.make_dummy_table_catalog(
+        rd_sca, bandpasses=[metadata['instrument']['optical_element']], nobj=10)
+
+    # Brighten sources
+    cat['F158'] = [flux * f158 for f158 in cat['F158']]
+
+    # Create catalog with one source for injection
+    source_cat = cat.copy()
+    source_cat.remove_rows(slice(0, 9))
+    xpos, ypos = twcs._xy(np.radians(source_cat['ra']), np.radians(source_cat['dec']))
```
schlafly, line 674: FWIW, I think I would have just given it an assigned position and type rather than something random, just because injecting a huge galaxy is not really what we want here.  e.g., a PSF.

schlafly (commented):
This looks good to me.  Do we have the option of making the image much smaller?  I prefer to avoid the 4k x 4k images in the regression suite as much as possible.

I think you should also make at least the injected source better defined than "the last of the random objects we generated", so that we can guarantee things like "the source lands on the image" and "the source is a PSF", which I think the later bits of the code assume?

Otherwise, looks good, thanks!

romanisim/tests/test_image.py
PaulHuwe: Both are implemented.

romanisim/tests/test_image.py
PaulHuwe: I reduced the size greatly. As part of that, I restricted the simulated objects to PSFs, to better fit the smaller image.

romanisim/tests/test_image.py
```diff
@@ -635,3 +635,109 @@ def test_reference_file_crds_match(level):
...
+    psfXmin = psf.image.bounds.getXMin() + 1
+    psfYmax = psf.image.bounds.getYMax() + 1
+    psfYmin = psf.image.bounds.getYMin() + 1
+
+    # Create injected source image
+    sourceobj = image.simulate_counts_generic(
+        sourcecounts, exptime, objlist=source_cat, psf=psf, zpflux=abflux, sky=None,
+        dark=None, flat=None,
+        bandpass=bandpass,
+        filter_name=filter_name, rng=rng)
```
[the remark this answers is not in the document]
PaulHuwe: I have changed this to utilize add_objects_to_image.

romanisim/tests/test_image.py
PaulHuwe: Injected source position is now specifically assigned.

romanisim/tests/test_image.py
```diff
@@ -635,3 +635,112 @@ def test_reference_file_crds_match(level):
...
+    ma_table = parameters.ma_table[metadata['exposure']['ma_table_number']]
+    tij = l1.ma_table_to_tij(ma_table)
+    tbar = ramp.ma_table_to_tbar(ma_table)
+    exptime = parameters.read_time * (ma_table[-1][0] + ma_table[-1][1] - 1)
+
+    # Create catalog of original sources
+    twcs = wcs.get_wcs(metadata, usecrds=False)
+    rd_sca = twcs.toWorld(galsim.PositionD(
+        galsim.roman.n_pix / 2, galsim.roman.n_pix / 2))
+    cat = table.vstack(catalog.make_stars(coord=rd_sca, radius=radius, rng=rng, n=nobj,
```
schlafly, line 670: This looks great.  What does the vstack do here?

schlafly (approved):
This looks great.  Feel free to go ahead and merge.  I left one minor comment; take it or leave it.  Then please update the test ticket to fill out more details of what you actually did.  We'll also need to make a release and upload new artifacts for the new build.

PR #92, opened 2023-11-28
Convert "ma_table" to read_pattern, and remove pkg_resources.

schlafly:
This PR removes references to "ma_table" in favor of "read_pattern", and removes the pkg_resources module in favor of importlib.resources.


schlafly:
@PaulHuwe , do you know more about importlib.resources than I do?  I did the simplest possible conversion from pkg_resources to importlib.resources, and did not introduce the context-managery bits with as_file discussed here:
https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename
Things seem to have worked anyway, but maybe there are some special egg modes that github & I aren't testing that would need the as_file treatment?

romanisim/bandpass.py
```diff
@@ -49,11 +49,17 @@ def read_gsfc_effarea(filename=None):
...
+                               header_start=1, data_start=2)
     if filename is None:
-        dirname = pkg_resources.resource_filename('romanisim', 'data')
-        filename = os.path.join(dirname, 'Roman_effarea_20201130.csv')
-    return Table.read(filename, format='csv', delimiter=',', header_start=1,
-                      data_start=2)
+        ref = (resources.files('romanisim') / 'data'
+               / 'Roman_effarea_20201130.csv')
+        with resources.as_file(ref) as filename:
+            out = reader(filename)
```
PaulHuwe, line 59: I am concerned about this breaking unexpectedly, due to the file being cleaned up with the `with` statement, as the reader is an iterator and not, say, a copy of the contents. I think we should use the `contextlib` stuff and control the lifetime of the temporary file.

schlafly: In this case reader is just Table.read(...)?  I would worry more if this were asdf with lazy loading, etc., but as far as I know that's not a concept with Table.read?

PaulHuwe: Hrm.. python won't close it until the `out` object is deleted, it just seems poor form to have a "hidden" attachment there, but this will still run fine (assuming nothing tries to clear the temporary space containing the file). Maybe I am simply being too conservative.

schlafly: The desired behavior here is that out is an in-memory object that has all of the file contents and resources.as_file(ref) never needs to be seen again; I expect the temporary file to be cleaned up when it exits the with statement.  I just don't expect to need this temporary file again after we've read its contents into memory in the out object.  But okay, I'm passing tests here and I'm following the idiom at https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename closely, so I think I'm going to declare victory here.

PR #93, opened 2023-12-01
Hotfix to Correct Source Injection Tests

PaulHuwe:
When the source injection test image was shrunk to 100x100 pixels, the previous tests weren't valid, as the PSF has a radius of 90 pixels. This ticket re-implements the spirit of the tests, as well as fixing a bug.

### DIFF
--- a/romanisim/tests/test_image.py
+++ b/romanisim/tests/test_image.py
@@ -729,13 +729,7 @@ def test_inject_source_into_image():
     assert np.any((origimage.data.value != newimage.value), where=nonzero)
 
     # Test that all pixels outside of the psf of the injected source are close to the original image
-    # Create array copies for all close comparison of non PSF-pixels
-    cleaned_origimage = copy.deepcopy(origimage.data.value)
-    cleaned_origimage[nonzero] = 1.0
-    cleaned_newimage = copy.deepcopy(newimage.value)
-    cleaned_newimage[nonzero] = 1.0
-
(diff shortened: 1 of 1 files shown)


romanisim/tests/test_image.py
```diff
@@ -726,14 +722,22 @@ def test_inject_source_into_image():
...
+    assert np.any((origimage.data.value != newimage.value), where=nonzero)
 
+    # Test that all pixels outside of the psf of the injected source are close to the original image
+    # Create array copies for all close comparison of non PSF-pixels
+    cleaned_origimage = copy.deepcopy(origimage.data.value)
+    cleaned_origimage[nonzero] = 1.0
+    cleaned_newimage = copy.deepcopy(newimage.value)
+    cleaned_newimage[nonzero] = 1.0
+
+    assert np.allclose(cleaned_origimage, cleaned_newimage, rtol=1e-05, atol=1e-08)
```
schlafly, line 738: Just trying to understand... is this equivalent to np.allclose(origimage.data.value[nonzero], newimage.data.value[nonzero])?

schlafly (approved):
Looks good.  I'm confused by a little of the logic but take a look and go ahead.

romanisim/tests/test_image.py
PaulHuwe: I have cleaned this up to something similar.

PR #96, opened 2024-02-07
Minor fix to clarify user-provided position angle instead of roll

Skyhawk172:
After some testing and discussion, it has become clear(er) that the user-provided ```--roll``` argument is really the position angle measured at the aperture's V2Ref/V3Ref (measured from North to YIdl). The handling of this value in ```wcs.py``` was also fixed for the case where the aperture used is BORESIGHT. 

Note that the code in ```wcs.py``` converts this value to ```roll_ref```, which is defined as the angle from North to V3 (towards East) measured at the aperture V2Ref/V3Ref, but from a user perspective, it makes more sense to specify the position angle of the aperture on the sky, hence the clarification provided here.

### DIFF
--- a/romanisim/wcs.py
+++ b/romanisim/wcs.py
@@ -38,7 +38,7 @@ except ImportError:
     import roman_datamodels.testing.utils as maker_utils
 
 
-def fill_in_parameters(parameters, coord, roll_ref=0, boresight=True):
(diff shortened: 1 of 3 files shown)


schlafly:
Great.  Thanks for catching both uses of roll in romanisim-make-image and romanisim-make-stack.  Merging!

PR #101, opened 2024-03-04
Add variable PSF option to romanisim and make it the default.

schlafly:
Adds support for a linear PSF variation across each SCA.

Formerly we assumed that the PSF was constant in each SCA.  This PR adds a new class that makes 4 PSFs at the 4 corners of each detector, and expresses the PSF at any point in the detector as a weighted linear sum of those 4 PSFs.  It also adds support in add_objects_to_image to use this new kind of PSF object.

There was an issue with ChromaticSums and convolutions and photon shooting in galsim 2.5.0; this also bumps the galsim requirement to 2.5.1 (the latest version).

This feature was motivated by the fact that romancal uses a gridded PSF that includes variation over each SCA, and astrometric residuals between romanisim and romancal showed best performance in the center of the SCA where the romanisim PSF was instantiated.  Using variable PSFs in both romancal and romanisim improves the astrometric agreement by a factor of three or so in F087.


schlafly:
@PaulHuwe , mind giving this a review if you're interested, or a LGTM?  The chief thing is adding point spread variation following the approach of roman_imsim, linearly interpolating among the four corners of a chip.  I also added a pin to pytest versions <= 8.0, since otherwise 8.1 was causing problems due to pytest plugins that do something with path that was deprecated, dunno.

romanisim/psf.py
```diff
@@ -118,3 +118,92 @@ def make_psf(sca, filter_name, wcs=None, webbpsf=True, pix=None,
...
+class VariablePSF:
+    """Spatially variable PSF wrapping GalSim profiles.
+
+    Linearly interpolates between four corner PSF profiles by summing
+    weighted GalSim PSF profiles.
+    """
+
+    def __init__(self, corners, psf):
+        self.corners = corners
+        self.psf = psf
```
PaulHuwe, line 179: Maybe labels these psfs, to retain the clarity that this is a set of psfs (rather than a single psf)?

PaulHuwe (approved):
LGTM with a tiny comment

PR #109, opened 2024-04-10
Allow SCA -1 to indicate that multiple images should be made.

schlafly:
This PR adds a few assorted romanisim features:
- specifying SCA -1 to romanisim-make-image means looping over all 18 detectors
- Use a seed for the flat field implementation rather than making a separate draw for the flat field, to enable deterministic output.
- Improve the handling of WCS in output products.
- Instantiate the PSFs a little further from the corners of the WFI detectors
- Handle NaNs in positions and fluxes in catalogs by ignoring these sources.


PaulHuwe (approved):
LGTM

PR #111, opened 2024-04-18
Truncate some too-long dummy metadata strings.

schlafly:
There are a handful of dummy strings in from the roman_datamodels maker utilities that are too long to be parsed by the archive.  We truncate those here to be be shorter.

We could consider making the defaults for these strings fit within the archive schema lengths in roman_datamodels or having validation code check these string lengths.  That would avoid this sort of issue.

### DIFF
--- a/romanisim/util.py
+++ b/romanisim/util.py
@@ -231,6 +231,20 @@ def add_more_metadata(metadata):
     target['proper_motion_epoch'] = 'J2000'
     target['source_type'] = 'EXTENDED'
 
+    # there are a few metadata keywords that have problematic, too-long
+    # defaults in RDM.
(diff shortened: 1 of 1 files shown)


PaulHuwe (approved):
Are you going to make an issue in RDM for this, or should I?

schlafly:
Is it clear to you that you would want to do this in RDM?  Ideally validate would do it, but I don't know if we have special RDM code in validate or if that just gets passed straight down to asdf, which doesn't know about these archive keywords?  You could just update the default values in RDM which would be easier, I guess.

PaulHuwe:
RDM wraps `validate`, so this could be added there.

PR #120, opened 2024-05-06
RCAL-789: Source Injection - Level 3

PaulHuwe:
Testing source injections into a level 3-like image.


romanisim/tests/test_image.py
```diff
@@ -746,3 +746,113 @@ def test_inject_source_into_image():
...
+    metadata['exposure']['ma_table_number'] = 1
+    bandpass = [metadata['instrument']['optical_element']]
+    filter_name = metadata['instrument']['optical_element']
+    sca = 4
+
+    # Create PSF
+    twcs = wcs.get_wcs(metadata, usecrds=False)
+    psf = romanisim.psf.make_psf(sca, filter_name, wcs=twcs, chromatic=False, webbpsf=True)
+
+    # Create initial Level 2-like image
```
schlafly, line 771: L3-like?

romanisim/tests/test_image.py
```diff
@@ -746,3 +746,113 @@ def test_inject_source_into_image():
...
+    la_arr = np.zeros([200, 200])
+
+    # Populate the image array with gaussian noise from generators
+    g1.generate(la_arr[0:100, 0:100])
+    g2.generate(la_arr[0:100, 100:200])
+    g3.generate(la_arr[100:200, 0:100])
+    g4.generate(la_arr[100:200, 100:200])
+
+    # Poisson Noise of image array
+    var_poisson = (la_arr - 1)**2
```
schlafly, line 793: I would make this image piecewise constant nad equal to 0.01^2, 0.02^2, 0.05^2, 0.1^2, and not have the "noise" component of the uncertainty, if that makes sense.  This is just for simplicity, really; in any case, we're just setting up an example image in this phase.

schlafly (commented):
This looks right to me.  But can you move the main simulation loop to a new file, say, l3.py, and call the function something like
add_objects_to_l3(l3, cat)
The l3 image should have the Poisson variance array needed to compute the Ct.  Return should be None probably indicating that the L3 image is updated in place.  Then the function basically executes the same ~loop you have in the code, looping over the objects in cat, computing Ct, doing the rendering, dividing, and adding to the image?  I guess you actually need two loops, one to compute all of the Ct, and one to render and add and divide, so that you don't get the wrong Ct because you changed the L3 image in the meantime.

Then this test would call that function.

What is "la" in "la_arr"?

Let's just make the test np.all(l3_new >= l3_old) and np.any(l3_new > l3_old) or something.  I think that's basically what your current test is doing.

For your test on lines 856 - 858, I worry I'm missing something.  I think we've injected 4 sources with the same flux in physical units.  We've injected them with different effective exposure times, so different amounts of noise.  But that noise can be positive or negative, so I would have guessed that some times you fail some of 856 - 858.  We should chat about that some time soon.

romanisim/tests/test_image.py
```diff
@@ -746,3 +747,86 @@ def test_inject_source_into_image():
...
+    # (total files contributed to each quadrant)
+
+    # Create gaussian noise generators
+    g1 = galsim.GaussianDeviate(31415926, mean=1.0, sigma=0.01)
+    g2 = galsim.GaussianDeviate(31415926, mean=1.0, sigma=0.02)
+    g3 = galsim.GaussianDeviate(31415926, mean=1.0, sigma=0.05)
+    g4 = galsim.GaussianDeviate(31415926, mean=1.0, sigma=0.1)
+
+    # Create level 3-like image model
+    l3_img = maker_utils.mk_level2_image(shape=(galsim.roman.n_pix, galsim.roman.n_pix))
```
schlafly, line 780: Could we do this with a real L3 model?

romanisim/tests/test_image.py
```diff
@@ -746,3 +747,86 @@ def test_inject_source_into_image():
...
+    # Set constants and metadata
+    galsim.roman.n_pix = 200
+    rng_seed = 42
+    metadata = copy.deepcopy(parameters.default_parameters_dictionary)
+    metadata['instrument']['detector'] = 'WFI07'
+    metadata['instrument']['optical_element'] = 'F158'
+    metadata['exposure']['ma_table_number'] = 1
+
+    # Create PSF
+    twcs = wcs.get_wcs(metadata, usecrds=False)
```
schlafly, line 765: This is a WCS, not a PSF.

romanisim/tests/test_wcs.py
```diff
@@ -13,10 +13,10 @@
...
 
 import roman_datamodels
-try:
-    import roman_datamodels.maker_utils as maker_utils
-except ImportError:
-    import roman_datamodels.testing.utils as maker_utils
+# try
+import roman_datamodels.maker_utils as maker_utils
+# except ImportError:
+#     import roman_datamodels.testing.utils as maker_utils
```
schlafly, line 19: Feel free to delete the old API.

schlafly (commented):
I think you didn't push the l3.py file?  I don't see it.

romanisim/tests/test_image.py
schlafly: (right now it's mk_level2_image rather than something with mosaic / l3 in it)

schlafly (approved):
Fine to merge when you're ready.

PR #123, opened 2024-05-17
Add a bounding box, but don't use it in romanisim.

schlafly:
@tddesjardins , care to see if this addresses your issue?

It's slightly annoying since by default the bounding box prevents one from figuring out, e.g., where stars land, if they land slightly off the detector.  So I add a box but then disable its use within the simulator.


tddesjardins:
LGTM. I think this will do the trick. Thanks!

PR #127, opened 2024-05-31
[SCSB-155] build with Numpy 2.0

zacharyburnett:
resolves [SCSB-155](https://jira.stsci.edu/browse/SCSB-155)

Numpy 2.0 is scheduled for release on June 19th. Due to the C ABI updates, Python projects that build C extensions are required to build against Numpy 2.0 in order to use Numpy 2.0 in runtime.
This is backwards-compatible with Numpy versions in runtime as far back as `1.26`

this PR pins `numpy>=2.0.0rc2` in `build-system.requires`
> [!NOTE]
> this PR was generated automatically by ``batchpr`` :robot:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ requires = [
     "setuptools_scm[toml] >=3.4",
     "wheel",
     "Cython >=0.29.21",
-    "numpy >=2.0.0rc2",
+    "numpy >=2.0.0",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ requires = [
     "setuptools_scm[toml] >=3.4",
     "wheel",
     "Cython >=0.29.21",
-    "numpy >=1.18",
+    "numpy >=2.0.0rc2",
 ]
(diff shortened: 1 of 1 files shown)


zacharyburnett:
converting this back to draft to address WebbPSF data issues in the CI

schlafly:
Oh, thanks for the reminder, I fixed the WebbPSF files here:
https://github.com/spacetelescope/romanisim/pull/126
I do have to clear the cache every time, so there is something weird about the hashing, but it only needs to be done a few times a year.

schlafly:
This was left as a draft, so I ignored it; should we move forward with this?

pyproject.toml
```diff
@@ -72,7 +72,7 @@ requires = [
     "setuptools_scm[toml] >=3.4",
     "wheel",
     "Cython >=0.29.21",
-    "numpy >=1.18",
+    "numpy >=2.0.0rc2",
```
zacharyburnett, line 75:
```suggestion
    "numpy >=2.0.0",
```

schlafly:
Sorry, just because I'm dumb; this isn't setting the dependency to require numpy > 2, but is instead setting something in the build system, which has little meaning to me.  What are we trying to accomplish here?

PR #128, opened 2024-06-04
abstract data download step into separate reusable workflow

zacharyburnett:
attempt to make WebbPSF data download / caching step simpler and more maintainable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,43 +26,9 @@ jobs:
       envs: |
         - linux: check-style
   data:
-    name: Download data
-    runs-on: ubuntu-latest
-    outputs:
-      path: ${{ steps.data.outputs.path }}
-      hash: ${{ steps.data_hash.outputs.hash }}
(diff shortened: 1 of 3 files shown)


schlafly:
This was left as a draft and I ignored it; should we move forward here?

zacharyburnett:
> This was left as a draft and I ignored it; should we move forward here?

I don't remember why I left it as a draft, but if it works and the tests pass then go ahead

zacharyburnett:
oh wait, the tests DON'T pass, that must be why. I'll take a look

schlafly:
I'm looking at some test failures in my unrelated PR, which is why I saw these again.  Maybe take another look tomorrow.

zacharyburnett:
I still had to delete the cache manually, as a newer data version is out now. We could fix that automatically by periodically downloading the data and comparing the hash to the existing cache, or just continue deleting the cache every so often

schlafly:
Looks good, merging!

PR #129, opened 2024-06-25
Document feature for generating images with all SCAs

bmorris3:
On main, users can generate images for all SCAs using `romanisim-make-image ... --sca -1 ...`, in a case that gets caught here:
https://github.com/spacetelescope/romanisim/blob/e0e6a979db8b43cbdb22e8082ae14e9ed2f8cd4d/scripts/romanisim-make-image#L27-L37

This feature isn't written about in the docs or in the argument's help notes, so I've added a note to both.


schlafly (approved):

schlafly:
Thank you!

PR #130, opened 2024-07-02
Add frame zero effects.

schlafly:
Adds correlated noise across all reads associated with noise in the reference read (frame zero), which is subtracted from all subsequent reads.

A new parameter, pedestal_extra_noise, is added, with a default value of 4 DN.  Noise is added to all resultants in each pixel with amplitude equal to the quadrature sum of the pixel's read noise and the extra pedestal noise.

This noise is degenerate with the pedestal and so doesn't meaningfully affect final ramp measurements at present.

We use a separate RNG so that pedestal noise can be turned on and off without affecting the rest of the random numbers in the simulation.

I also added corresponding tests.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -756,6 +756,7 @@ def simulate(metadata, objlist,
     saturation = refdata['saturation']
     reffiles = refdata['reffiles']
     flat = refdata['flat']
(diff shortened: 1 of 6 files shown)


schlafly:
@PaulHuwe , want to take a look?

PaulHuwe (approved):
LGTM

PR #132, opened 2024-07-07
RCAL-862 Refactor L3 Injection

PaulHuwe:
This PR greatly reduces the code (and loop) duplication of level 3 source injection. Updates were made to the image source injection and the test. The l3 test was moved from test_image to test l3. Support code was added to wcs and parameters.

NOTE: The l3 and l3_test import blocks are large due to other PRs happening simultaneously.


romanisim/image.py
```diff
@@ -217,33 +214,38 @@ def trim_objlist(objlist, image):
 
 
 def add_objects_to_image(image, objlist, xpos, ypos, psf,
-                         flux_to_counts_factor, bandpass=None, filter_name=None,
-                         rng=None, seed=None):
+                         flux_to_counts_factor, exptimes=None,
```
schlafly, line 217: A few thoughts on this function:
- You've looked more closely than I have, but there's some kind of ~duplication or redundancy about exptimes and flux_to_counts_factor.  For L2 operation, you just need one of these.  For L3 operation, we scale the flux_to_counts_factor by exptime and then divide it out by exptime later.  I can imagine something like scale_exptime instead, that defaults to 1, and multiplies flux_to_counts_factor and is also used as a denominator.  Not needed as a change, but worth thinking about.
- If we kept the API to only support galsim objects, would that make the caller significantly more complicated?  That would make the API simpler and not need to have the WCS come in in two places.

romanisim/l3.py
```diff
@@ -17,64 +36,56 @@ def add_objects_to_l3(l3_mos, source_cat, rng=None, seed=None):
...
+
+    # Create position arrays (if needed)
+    if any(pos is None for pos in [xpos, ypos]):
+        # Create coordinates (if needed)
+        if coords is None:
+            coords = np.array([[o.sky_pos.ra.rad, o.sky_pos.dec.rad]
+                               for o in source_cat])
+            coords_unit = 'rad'
+        # Generate x,y positions for sources
+        sourcecountsall = galsim.ImageF(l3_mos.data.shape[0], l3_mos.data.shape[1], wcs=wcs, xmin=0, ymin=0)
```
schlafly, line 81: This image doesn't appear to be needed?

romanisim/l3.py
```diff
@@ -17,64 +36,56 @@ def add_objects_to_l3(l3_mos, source_cat, rng=None, seed=None):
...
+        xpos, ypos = sourcecountsall.wcs.radecToxy(coords[:, 0], coords[:, 1], coords_unit)
+
+    # Add sources to the original mosaic data array
+    romanisim.image.add_objects_to_image(l3_mos.data, source_cat, xpos=xpos, ypos=ypos,
+                                         psf=psf, flux_to_counts_factor=exptimes, exptimes=exptimes,
```
schlafly, line 86: I think the flux_to_counts_factor needs to be multiplied by the AB flux zero point for this filter?  i.e., add_objects_to_image needs to know how many counts to put in, but the fluxes are specified in maggies, so the flux_to_counts_factor needs to know the maggie -> photon relation as well as the exposure time.  You may be doing something related in the photom section of the parameter file, but I didn't see how that connected.

romanisim/parameters.py
```diff
@@ -84,6 +84,15 @@
...
     "saturation": 55000 * u.DN,
+    "photom": {"F062": 0.28798833 * u.MJy / u.sr,
+               "F087": 0.3911463 * u.MJy / u.sr,
+               "F106": 0.37039083 * u.MJy / u.sr,
+               "F129": 0.36495176 * u.MJy / u.sr,
+               "F146": 0.11750617 * u.MJy / u.sr,
+               "F158": 0.35352992 * u.MJy / u.sr,
+               "F184": 0.53612772 * u.MJy / u.sr,
+               "F213": 0.55763922 * u.MJy / u.sr,
+               }
```
schlafly, line 95: I missed where this is used.

romanisim/tests/test_l3.py
```diff
@@ -0,0 +1,133 @@
...
+    # Set locations
+    xpos_idx = [50, 50, 150, 150]
+    ypos_idx = [50, 150, 50, 150]
+
+    # Populate flux scaling ratio and catalog
+    Ct = []
+    for idx, (x, y) in enumerate(zip(xpos_idx, ypos_idx)):
+        # Set scaling factor for injected sources
+        # Flux / sigma_p^2
+        if l3_mos.var_poisson[x][y].value != 0:
```
schlafly, line 88: I think I at least ended up thinking we needed [y, x] here.

romanisim/image.py
```diff
@@ -289,21 +295,35 @@ def add_objects_to_image(image, objlist, xpos, ypos, psf,
...
         if chromatic:
             stamp = final.drawImage(
-                bandpass, center=image_pos, wcs=image.wcs.local(image_pos),
+                bandpass, center=image_pos, wcs=pwcs,
                 method='phot', rng=rng)
         else:
             try:
                 stamp = final.drawImage(center=image_pos,
-                                        wcs=image.wcs.local(image_pos))
+                                        wcs=pwcs)
```
schlafly, line 309: Probably around here need a:
if add_noise:
  stamp.addNoise(galsim.PoissonNoise(rng))
to get around that issue we were discussing that noise is only added in photon-shooting / chromatic mode by default.

romanisim/image.py
PaulHuwe: I separated them to allow for unit change, if desired.. I could rename them to make that more clear?

We can have it only support galsim objects, it just requires making one of the base mosaic that objects are being written to within the L3 code - I can make that change.

romanisim/l3.py
PaulHuwe: Isn't it needed to translate RA & DEC to X & Y?

romanisim/parameters.py
PaulHuwe: Sorry, this is used in other L3 code and came along for the ride here.

romanisim/l3.py
schlafly: Sorry, I landed on the wrong line.  My point is that I think you make a galsim image sourcecountsall and then don't use it.  OTOH I suggested above to actually use it and so to do add_objects_to_image(sourcecountsall, ...) below.  But as written I don't think you really use sourcecountsall.

romanisim/tests/test_l3.py
PaulHuwe: Corrected.

romanisim/image.py
PaulHuwe: I was saving that for RCAL-863.

romanisim/l3.py
PaulHuwe: You are correct on my using the photom values for that, but I had it outside of add_objects_to_image. 
https://github.com/PaulHuwe/romanisim/blob/RCAL-832_L3Generation/romanisim/l3.py#L361
I can move it in.

romanisim/image.py
PaulHuwe: Conversion to image implemented.

PaulHuwe: I didn't rename the variables, though I did include the unit scaling you mentioned below. I am willing to discuss the variable names /usage further, if you desire.

romanisim/image.py
```diff
@@ -227,23 +225,27 @@ def add_objects_to_image(image, objlist, xpos, ypos, psf,
...
+    exptimes: array_like
+        Exposure times to scale back to rate units
     bandpass : galsim.Bandpass
         bandpass in which image is being rendered.  This is used only in cases
         where chromatic profiles & PSFs are being used.
     filter_name : str
         filter to use to select appropriate flux from objlist.  This is only
         used when achromatic PSFs and sources are being rendered.
+    wcs : galsim.GSFitsWCS
+        WCS corresponding to image
```
schlafly, line 248: You can now remove the wcs argument, and always use image.wcs?

romanisim/image.py
```diff
@@ -289,18 +295,24 @@ def add_objects_to_image(image, objlist, xpos, ypos, psf,
             psf0 = psf.at_position(xpos[i], ypos[i])
         else:
             psf0 = psf
-        final = galsim.Convolve(profile * flux_to_counts_factor, psf0)
+        if isinstance(flux_to_counts_factor, list):
+            final = galsim.Convolve(profile * flux_to_counts_factor[i], psf0)
+        else:
+            final = galsim.Convolve(profile * flux_to_counts_factor, psf0)
```
schlafly, line 301: Minor stylistic thing, but my preference would be to define something like factor = flux_to_counts_factor[i] if isinstance(...) else flux_to_counts_factor and then to let the convolution like be galsim.Convolve(profile * factor, psf0), just aiming to avoid duplicating the convolution line.

romanisim/l3.py
```diff
@@ -1,14 +1,33 @@
...
+import romanisim.bandpass
+import romanisim.psf
+import romanisim.image
+import romanisim.persistence
+from romanisim import log
+import roman_datamodels.maker_utils as maker_utils
 
-def add_objects_to_l3(l3_mos, source_cat, rng=None, seed=None):
+# Define centermost SCA for PSFs
+CENTER_SCA = 2
```
schlafly, line 26: Let's probably put this in romanisim.parameters, possibly renaming it to default_sca as there isn't a good center.

romanisim/l3.py
```diff
@@ -17,64 +36,63 @@ def add_objects_to_l3(l3_mos, source_cat, rng=None, seed=None):
...
+                               for o in source_cat])
+            coords_unit = 'rad'
+        # Generate x,y positions for sources
+        xpos, ypos = sourcecountsall.wcs.radecToxy(coords[:, 0], coords[:, 1], coords_unit)
+
+    # Add sources to the original mosaic data array
+    romanisim.image.add_objects_to_image(sourcecountsall, source_cat, xpos=xpos, ypos=ypos,
+                                         psf=psf, flux_to_counts_factor=[xpt * unit_factor for xpt in exptimes],
+                                         exptimes=exptimes, bandpass=[filter_name], filter_name=filter_name,
+                                         wcs=wcs, rng=rng, seed=seed)
```
schlafly, line 92: Our naming isn't quite right, but the flux_to_counts_factor needs to take the source catalog fluxes (maggies in the simulator) into counts.  That factor will be the AB flux for the particular filter times the exposure time; that's close to what you have.  Then the "exptimes" factor is what we're using to go back from counts to whatever units the L3 file is in; MJy / sr.  That factor needs to be:
abflux * 3631 * 10^6 / (pix / steradian)
if I got it all right.  abflux puts one back in maggies / pixel units.  A maggie is the same as 3631 Jy, so then one needs to divide by 3631, convert from Jy to MJy, and and convert from pixels to steradians.  Let's just use a pixel in the center of the mosaic, though an argument could be made for computing the size of each pixel individually.

romanisim/parameters.py
schlafly: And conceptually, this is the number of MJy / sr that is equal to 1 count / s in a fiducial L2 pixel?  Hopefully that's the same as (count / s / pix) / Jy * 10^6 Jy / MJy * pix / (0.1 / 60 / 60 * 180 / pi)^2 = abflux / 3631 * 10^6 Jy / MJy * pix / (0.1 / 60 / 60 * 180 / pi)^2?

romanisim/l3.py
PaulHuwe: Changed

romanisim/image.py
PaulHuwe: I have removed it, though since l3_mos can be a datamodel, I had to tweak the code a little.

PaulHuwe: I have removed this from image and not from l3

romanisim/image.py
PaulHuwe: Implemented.

romanisim/image.py
PaulHuwe: I renamed one of the variables as part of a change below.

romanisim/l3.py
PaulHuwe: Implemented.

romanisim/parameters.py
```diff
@@ -84,14 +84,25 @@
...
+                            },
+               "pixelareasr": {"F062": 2.640600009241359e-13 * u.sr,
+                               "F087": 2.640600009241359e-13 * u.sr,
+                               "F106": 2.640600009241359e-13 * u.sr,
+                               "F129": 2.640600009241359e-13 * u.sr,
+                               "F146": 2.640600009241359e-13 * u.sr,
+                               "F158": 2.640600009241359e-13 * u.sr,
+                               "F184": 2.640600009241359e-13 * u.sr,
+                               "F213": 2.640600009241359e-13 * u.sr,
+                               }
```
schlafly, line 105: I thought the decision here was to remove this in favor of get_abflux()

romanisim/l3.py
```diff
@@ -87,12 +89,14 @@ def add_objects_to_l3(l3_mos, source_cat, exptimes, xpos=None, ypos=None, coords
 
     # Add sources to the original mosaic data array
     romanisim.image.add_objects_to_image(sourcecountsall, source_cat, xpos=xpos, ypos=ypos,
-                                         psf=psf, flux_to_counts_factor=[xpt * unit_factor for xpt in exptimes],
-                                         exptimes=exptimes, bandpass=[filter_name], filter_name=filter_name,
-                                         wcs=wcs, rng=rng, seed=seed)
+                                         psf=psf, flux_to_counts_factor=exptimes,
```
schlafly, line 92: flux_to_counts_factor still needs the factor of abflux?

schlafly (commented):
This looks good.  I think the flux_to_counts_factor is missing the factor of abflux now, though; we should reflect a bit about units.

Since they're providing a source_cat the input units are ~required to be in maggies.  The ultimate flux_to_counts_factor needs to then be abflux * exptime.  Then there's the denominator, now convtimes, that needs to be counts / (target units), so 1 / exptime / (MJy / sr / (count / s)).

The check should probably be that the sum of the output image times the pixel size divided by 3631 is equal to the flux after conversion to Janskies, if that makes sense?

romanisim/parameters.py
PaulHuwe: removed

romanisim/l3.py
PaulHuwe: Should all be addressed.

romanisim/parameters.py
```diff
@@ -84,6 +84,17 @@
...
     "saturation": 55000 * u.DN,
+    # Taken from roman_wfi_photom_0046.asdf
+    "photom": {"pixelareasr": {"F062": 2.640600009241359e-13 * u.sr,
+                               "F087": 2.640600009241359e-13 * u.sr,
+                               "F106": 2.640600009241359e-13 * u.sr,
+                               "F129": 2.640600009241359e-13 * u.sr,
+                               "F146": 2.640600009241359e-13 * u.sr,
+                               "F158": 2.640600009241359e-13 * u.sr,
+                               "F184": 2.640600009241359e-13 * u.sr,
+                               "F213": 2.640600009241359e-13 * u.sr,
```
schlafly, line 95: This looks like it's only used in the tests; let's move it there if we need this?  I think my preference would be that users mostly use the WCS to figure out the pixel scale.

romanisim/wcs.py
```diff
@@ -431,27 +432,34 @@ def get_mosaic_wcs(mosaic):
...
     # Create a tangent plane WCS for the mosaic
     # The affine parameters below should be reviewed and updated
     affine = galsim.AffineTransform(
-        0.1, 0, 0, 0.1, origin=galsim.PositionI(mosaic_node.data.shape[0] / 2.0,
-                                                mosaic_node.data.shape[1] / 2.0,),
+        0.1, 0, 0, 0.1, origin=galsim.PositionI(math.ceil(shape[0] / 2.0), math.ceil(shape[1] / 2.0)),
         world_origin=galsim.PositionD(0, 0))
     wcs = galsim.TanWCS(affine,
                         util.celestialcoord(world_pos))
+    # wcs = GWCS(wcs)
```
schlafly, line 463: Let's remove this.

schlafly (approved):
Some very minor suggestions inline, but good to go, thanks!

romanisim/parameters.py
PaulHuwe: This is used in other L3 code, so should stay.

PR #133, opened 2024-07-16
Add support for image input

schlafly:
Add support for image input to the simulator.

Conceptually, the idea is that users may have a set of "truth" data as to what real galaxies look like.  This could come from hydro simulations or deep images of galaxies.  These galaxies have better fidelity than the kind of simple parametric galaxies ordinary catalog input supports.

This PR adds a new kind of catalog input mode, where the catalog input includes an 'ident' field that indicates which galaxy in a catalog of galaxy images should be input into an image at a given location.  On top of that, it includes additional fields that specify how that galaxy image should be sheared, dilated, or rotated.  The image is then rendered with the given galaxies at the given locations.

This feature leverages the RealGalaxyCatalog class in GalSim, allowing it to be used in combination with existing true galaxies catalogs in the GalSim format, though only a subset of the GalSim features are presently supported.

A new function `catalog.make_image_catalog` has been added to make it easier to make new catalogs of true images, and this feature is exercised in the unit tests to make some trivial disc and rectangle galaxies.

Caveat emptor: the RealGalaxyCatalog relies on the images to be simulated to have some point spread function.  This point spread function is then deconvolved from the image before the image is convolved with the Roman PSF.  This leads to ringing when the Roman PSF is sharper than the original image PSF---essentially, the simulation is being asked to invent sub-pixel information, and it does not do so well.  This also occurs if the image is dilated significantly before rendering: the dilation means that the same input pixels correspond to more output pixels, and again information must be invented.

This PR implements DMS 228 and adds construction of the corresponding artifacts.

### DIFF
--- a/docs/romanisim/image-input.rst
+++ b/docs/romanisim/image-input.rst
@@ -1,7 +1,7 @@
 Image Input
 ===========
 
-The simulator can alternatively take as input a catalog that references a separate list of images of sources.  The format of the catalog is described below, and the underlying "list of images" is described by a ``RealGalaxyCatalog``; see :ref:`Constructing a RealGalaxyCatalog`.  The concept here is that users may have noiseless images of galaxies from, for example, hydro simulations, and may wish to inject those into simulated images at specific locations.  Or users may have reference images from, for example, the COSMOS field, which could be used as truth images for simulations.
+The simulator can alternatively take as input a catalog that references a separate list of images of sources.  The format of the catalog is described below, and the underlying "list of images" is described by a ``RealGalaxyCatalog``; see :ref:`constructing-a-rgc`.  The concept here is that users may have noiseless images of galaxies from, for example, hydro simulations, and may wish to inject those into simulated images at specific locations.  Or users may have reference images from, for example, the COSMOS field, which could be used as truth images for simulations.
 
 The simulator leverages GalSim's ``RealGalaxyCatalog`` framework to support this use case.  There are a handful of existing catalogs of images in the appropriate format for use with the ``RealGalaxyCatalog``---in particular, `GalSim's COSMOS reference catalog <https://galsim-developers.github.io/GalSim/_build/html/real_gal.html#downloading-the-cosmos-catalog>`_ .
 
@@ -42,6 +42,7 @@ Following these required fields is a series of columns giving the fluxes of the 
 
 The simulator then renders these images in the scene and produces the simulated L1 or L2 images.
 
+.. _constructing-a-rgc:
 
 Constructing a RealGalaxyCatalog
 ================================
--- a/docs/romanisim/image-input.rst
+++ b/docs/romanisim/image-input.rst
@@ -1,0 +1,52 @@
+Image Input
+===========
+
+The simulator can alternatively take as input a catalog that references a separate list of images of sources.  The concept here is that users may have noiseless images of galaxies from, for example, hydro simulations, and may wish to inject those into simulated images at specific locations.  Or users may have reference images from, for example, the COSMOS field, which could be used as truth images for simulations.
+
+The simulator leverages GalSim's ``RealGalaxyCatalog`` framework to support this use case.  There are a handful of existing catalogs of images in the appropriate format for use with the ``RealGalaxyCatalog``---in particular, `GalSim's COSMOS reference catalog <https://galsim-developers.github.io/GalSim/_build/html/real_gal.html#downloading-the-cosmos-catalog>`_ .
+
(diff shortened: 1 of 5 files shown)


schlafly:
@PaulHuwe , this should be ready; feel free to review at your leisure.  Thanks!

romanisim/tests/test_image.py
```diff
@@ -847,3 +822,81 @@ def test_inject_sources_into_mosaic():
...
+    im1 = ((xx ** 2 + yy ** 2) < 30 ** 2) * 1.0  # circle
+    im2 = im1 * 0
+    im2[35:65, 10:-10] = 1  # rectangle
+
+    # make a PSF for these
+    psf = im1 * 0
+    psf[cenpix, cenpix] = 1
+    from scipy.ndimage import gaussian_filter
+    sigma = 1
+    psf = gaussian_filter(psf, 1)
```
PaulHuwe, line 843:
```suggestion
    psf = gaussian_filter(psf, sigma)
```

docs/romanisim/image-input.rst
```diff
@@ -0,0 +1,52 @@
+Image Input
+===========
+
+The simulator can alternatively take as input a catalog that references a separate list of images of sources.  The concept here is that users may have noiseless images of galaxies from, for example, hydro simulations, and may wish to inject those into simulated images at specific locations.  Or users may have reference images from, for example, the COSMOS field, which could be used as truth images for simulations.
```
PaulHuwe, line 4: This section is a bit confusing when it comes to what the total inputs are.. maybe a sentence with a list? I like the mention of the test to show how to work through this.

PaulHuwe (changes requested):
One additional comment.. it can be confusing to read image(s) used throughout to refer to either the scene or the object postage stamps. It might be helpful to change that verbiage?

romanisim/catalog.py
```diff
@@ -334,6 +335,140 @@ def make_stars(coord,
...
+    Returns
+    -------
+    None - this routine only writes files to out_base_filename
+    """
+    outdtype = [('ident', 'i4'), ('gal_filename', 'U200'), ('psf_filename', 'U200'),
+                ('noise_file_name', 'U200'), ('gal_hdu', 'i4'), ('psf_hdu', 'i4'),
+                ('pixel_scale', 'f4'), ('noise_variance', 'f4'), ('mag', 'f4'),
+                ('band', 'U10'), ('weight', 'f4'), ('stamp_flux', 'f4')]
+    nimage = len(image_filenames)
+    res = np.zeros(nimage, dtype=outdtype)
```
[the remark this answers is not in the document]
schlafly: Sorry, res is the output catalog I eventually write out; it's a structured array?

romanisim/catalog.py
```diff
@@ -334,6 +335,140 @@ def make_stars(coord,
...
+                       pixel_scale=0.05):
+    """Construct a RealGalaxyCatalog from a list of image filenames.
+
+    GalSim supports catalogs of real galaxies from images that can be inserted
+    into images.  This function makes it easy to produce a minimal set of files
+    that can be used as a GalSim RealGalaxyCatalog from list list of fits
+    images.  These images can come from anywhere, but are expected to come from
+    either real imaging or hydro simulations.
+
+    This routine assumes that all images have a pixel scale of 0.05" and are
```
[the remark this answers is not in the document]
schlafly: I actually deleted this since now 0.05" is just the default and other options are supported, thanks.

romanisim/tests/test_image.py
schlafly: Fixed, thanks.

schlafly:
Sorry @zacharyburnett , but something is going wrong with the WebbPSF download again... would you mind taking a look?  https://github.com/spacetelescope/romanisim/actions/runs/10047465457/job/27804297949?pr=133

zacharyburnett:
> Sorry @zacharyburnett , but something is going wrong with the WebbPSF download again... would you mind taking a look? https://github.com/spacetelescope/romanisim/actions/runs/10047465457/job/27804297949?pr=133

Sorry, I set up the `data.yml` workflow to run weekly but never actually manually ran it to remake the cache 😅 reran it, and that seems to have fixed the issue now

schlafly:
Thank you!

docs/romanisim/image-input.rst
schlafly: I've changed this to try to more consistently distinguish "input images" from "output images", but I agree that it's confusing.

romanisim/catalog.py
```diff
@@ -409,16 +409,15 @@ def make_image_catalog(image_filenames, psf, out_base_filename,
...
-    seen through a common PSF, which is given as the PSF argument.  This PSF is
-    deconvolved before reconvolving with the appropriate filter-specific PSF
-    when rendering a Roman image.
+    GalSim supports catalogs of real galaxies from input images that can be
+    inserted into output images.  This function makes it easy to produce a
+    set of files that can be used as a GalSim RealGalaxyCatalog from a list of
+    fits input images.  These input images can come from anywhere, but are
+    expected to come from either real imaging or from hydro simulations.
+
+    This routine assumes that all images a common PSF, which is given as the
```
PaulHuwe, line 418:
```suggestion
    This routine assumes that all images share a common PSF, which is given as the
```

PaulHuwe (approved):
One small tweak, otherwise, LGTM

PR #137, opened 2024-08-07
RCAL-832 L3 Generation

PaulHuwe:
This ticket adds Level 3 Mosaic simulation and tests.


PaulHuwe:
@zacharyburnett I have been trying to debug the failures above, to no avail. Is this something you can shed some light one?

zacharyburnett:
ah, both `build.yml` and the ReadTheDocs build are failing because importing `romanisim` checks for `WEBBPSF_PATH` which is not set. Could we remove that check / only check if modules are actually used?

romanisim/tests/test_l3.py
```diff
@@ -142,3 +143,364 @@ def test_inject_sources_into_mosaic():
...
+    im2 = im.copy()
+    im2, extras2 = l3.simulate_cps(im2, metadata, exptime, objlist=graycat,
+                                   xpos=[50] * len(chromcat), ypos=[50] * len(chromcat),
+                                   psf=imdict['impsfgray'], seed=rng_seed,
+                                   ignore_distant_sources=100)
+
+    # Ensure that the two simualtions are in line with each other
+    maxim = np.where(im1.array**2 > im2.array**2, im1.array, im2.array)
+    m = np.abs(im1.array**2 - im2.array**2) <= 20**2 * np.abs(maxim)
+    assert np.all(m)
```
schlafly, line 328: This looks like a huge, huge range?  You probably want abs(im1 - im2) < something, but 400 times the maximum of the image is weird?  But it's the difference of squares, so the units are different?  I don't think we actually need to compare chromatic vs. filter in the first place.

romanisim/tests/test_l3.py
```diff
@@ -142,3 +143,364 @@ def test_inject_sources_into_mosaic():
...
+                               psf=imdict['impsfgray'],
+                               cps_conv=1, unit_factor=(1 * u.MJy / u.sr), seed=rng_seed,
+                               metadata=metadata, sky=0,
+                               ignore_distant_sources=100,
+                               effreadnoise=0,
+                               )
+
+    # Ensure that the two simualtions are in line with each other
+    maxim = np.where(im3['data'].value**2 > im4['data'].value**2, im3['data'].value, im4['data'].value)
+    m = np.abs(im3['data'].value**2 - im4['data'].value**2) <= 20**2 * np.abs(maxim)
```
schlafly, line 349: Likewise.

romanisim/parameters.py
```diff
@@ -4,6 +4,9 @@
 import numpy as np
 from astropy.time import Time
 from astropy import units as u
+import webbpsf as wpsf
+
+pixel_scale = wpsf.WFI().pixelscale
```
schlafly, line 9: I think Zach is saying that if you replace these two lines with pixel_scale = 0.11 then your build issues will go away.  That sounds fine.

schlafly (commented):
I left a few comments inline.  I think I would delete the two tests involving comparisons against 20**2 * max(image), and I think you should get the build working, probably by removing the pixel_scale bit that depends on webbpsf.  Don't worry about the other comments; I'll take it from here.

romanisim/parameters.py
PaulHuwe: Thanks for your help, folks. Changed.

schlafly:
Great, thanks.  Can you also confirm what is intended for the L3 test with the 20**2, and then let's go ahead and merge?

schlafly (approved):

zacharyburnett:
just one warning when building the docs now:
```
... romanisim/wcs.py:docstring of romanisim.wcs.get_mosaic_wcs:5: ERROR: Unexpected indentation.
```

romanisim/tests/test_l3.py
PaulHuwe: Removed this as the sim vs sim_cps test is the important one.

romanisim/tests/test_l3.py
PaulHuwe: Removed this as the sim vs sim_cps test is the important one.

romanisim/l3.py
```diff
@@ -83,15 +99,501 @@ def add_objects_to_l3(l3_mos, source_cat, exptimes, xpos=None, ypos=None, coords
...
+        bandpass = roman.getBandpasses(AB_zeropoint=True)[galsim_filter_name]
+
+    # Create initial galsim image (x & y are flipped)
+    image = galsim.ImageF(shape[1], shape[0], wcs=wcs, xmin=0, ymin=0)
+
+    # Create sky for this mosaic, if not provided (in cps)
+    if sky is None:
+        date = meta['basic']['time_mean_mjd']
+        if not isinstance(date, astropy.time.Time):
+            date = astropy.time.Time(date, format='mjd')
```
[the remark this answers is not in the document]
PaulHuwe: Added

PR #139, opened 2024-08-12
[SCSB-170] pin Python<3.13

zacharyburnett:
Resolves [SCSB-170](https://jira.stsci.edu/browse/SCSB-170)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,12 +13,12 @@ jobs:
       env: |
         FFTW_DIR: /opt/homebrew/opt/fftw/lib/
       targets: |
-        # Linux wheels
-        - cp3*-manylinux_x86_64
-        # MacOS wheels
-        - cp3*-macosx_x86_64
-        # Until we have arm64 runners, we can't automatically test arm64 wheels
-        - cp3*-macosx_arm64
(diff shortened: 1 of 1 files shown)


schlafly:
I'm merging this since it resolves the build failures, but I'm confused about the content of the PR, which looks more like "build wheels for python != 3.13" and the title of the PR (pin python <3.13).

schlafly (approved):

schlafly:
Sounds good, thanks.  Do you understand the issue?  Somehow I expect python 3.13 problems to manifest as syntax errors or unit test errors and not only in compiling the binaries, but maybe if cython isn't happy with 3.13 then it might only appear when building wheels.

zacharyburnett:
> Sounds good, thanks. Do you understand the issue? Somehow I expect python 3.13 problems to manifest as syntax errors or unit test errors and not only in compiling the binaries, but maybe if cython isn't happy with 3.13 then it might only appear when building wheels.

I don't understand the issue, but from further experimentation with `cibuildwheel` yesterday I think we should revert this change and instead pin Python in `pyproject.toml` until we can figure out why it's failing; that will be much more maintainable and also prevent users from trying to install on 3.13 and getting installation errors. 

I'll make another PR

PR #141, opened 2024-08-13
[SCSB-170] pin Python<3.13 (take 2)

zacharyburnett:
Resolves [SCSB-170](https://jira.stsci.edu/browse/SCSB-170)

follow-on to #139


zacharyburnett:
if the `build.yml` checks pass then this should be good to merge

zacharyburnett:
@schlafly sorry for the second PR! I should have kept the first one in a draft until I was done experimenting

schlafly:
Thanks, I'm just glad when the red xs go away!

PR #143, opened 2024-08-26
Update decorator to include DMS223

schlafly:
I had omitted to include DMS223 in the metrics logger decorator.  This PR adds that missing decoration.


PaulHuwe (approved):
LGTM

PR #144, opened 2024-08-30
Improve APT program generation interface.

schlafly:
This PR adds a handful of things to better support generating APT programs with romanisim.

* It removes the old apt interface which we have never really used.
* It changes the filenames to use the new scheme that omits the group, sequence, and activity bits.
* It reads different metadata out of the APT file, aiming to be able to fill out PI / category / subcategory / ... information, but it doesn't yet pass that information in a sensible way forward when spitting out a script file.
* It adds a new merge_dicts utility function that does a recursive dictionary merger, so that we can update keys from APT into the main metadata dictionary.
* It adds a new script to generate Gaia catalogs for a region.
* It tweaks romanisim-make-stack somewhat to use the new APT output format (which includes plan/pass/segment/observation/visit information) and the new metadata grabber from APT.


romanisim/util.py
```diff
@@ -544,3 +544,31 @@ def update_photom_keywords(im, gain=None):
         0 * u.MJy / u.sr)
     im['meta']['photometry']['conversion_microjanskys_uncertainty'] = (
         0 * u.uJy / u.arcsec ** 2)
+
+
+def merge_dicts(a, b):
```
PaulHuwe, line 549: Should we update the code to use this in other places?

scripts/romanisim-make-stack
```diff
@@ -92,8 +85,8 @@ def main():
     # SCA detector identifier number
     parser.add_argument('-s', '--sca',
                         type=int,
-                        default=0,
-                        help='SCA to simulate')
+                        default=-1,
```
PaulHuwe, line 88: Did you ensure that this behaved as expected for both a specified SCA and for all?

PaulHuwe (approved):
We should add tests of parse_apt_file, ramanisim-gaia-catalog, romanisim-make-catalog, romanisim-make-image, and romanisim-make-stack.

romanisim/util.py
schlafly: We should, but I haven't done that yet.

scripts/romanisim-make-stack
schlafly: I only tested the "commands" mode, but this works for both single and multiple SCAs at present.

PR #156, opened 2024-09-17
More doc updates

### LINKED ISSUE #152: Roman I-Sim: readthedocs report
_Issue [RCAL-904](https://jira.stsci.edu/browse/RCAL-904) was created on JIRA by [Andrea Bellini](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bellini):_

We went through all pages in the readthedocs documentation. We list below our comments:

Main general comments:
 
— Overall, there is quite a bit of confusion throughout the documentation related to units. There is about a factor of 2 difference between electrons in the pixels/detectors and DNs in the read arrays, but both are usually referred to as ‘counts, which is a very generic term. Also, sometimes the word ‘photon(s)’ seems to be used instead of ‘electron(s)’, which make readers think the sentence is not describing the detector but the source. Please replace relevant instances of “photon(s)” and “count(s)” with the correct units (either electron(s) or DN(s)) according to the context. This will help clarify a lot of confusing statements.
 
— Also, there are a few pages, mostly related to the description of L1 and L2 images, containing what appears to be a list of personal notes and thoughts rather than a description/documentation of what the relevant functions/methods do. We strongly suggest to rewrite those pages in the spirit of a documentation style, and avoid the use of “I” or “we” subjects as much as possible.
 
— Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y], etc. This happens in many pages.
 
— The “Search docs” bar does not work. This might not be a romanisim-only issue, as the issue appears to be spread out to most/all readthedocs pages, but it should still be fixed.
 
 
Other general comments:
 
— In some pages equations are given using a latex interpreter while in others are simply typed down in ASCII, which are less easy to read. We suggest to always use the latex interpreter for clarity.
 
— Sometimes GalSim is rendered as ‘GalSim’, other as ‘galsim’ (even in the same page, e.g., <https://romanisim.readthedocs.io/en/latest/romanisim/wcs.html).> Also, WebbPSF is always rendered as ‘webbpsf’. Can these packages be consistently rendered throughout the documentation? If a distinction is made to distinguish between GalSim as the name of the program, and galsim as the python package, can a different font be used for python packages, to improve clarity?
 
 
Below we provide comments on a page-by-page basis and in order of appearance. Most are suggestions for text replacement to improve clarity.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/overview.html>
1. Replace 'convolution of the sources by the Roman point spread function' with 'convolution of the sources by the Roman point spread function (PSF)'. Reason: PSF is used in the same page but it is not defined.
2. Replace 'are using “official” webbpsf PSF' with 'are using “official” webbpsf PSFs". Reason: verb is plural and object is singular.
3. Replace 'and reference files from the Roman CRDS (not yet public!).' with 'and most reference files from the Roman CRDS.', where CRDS is a link to <https://roman-crds.stsci.edu/.> Reason: Roman CRDS is now public, and romanisim does not use all reference files in there.
4. Replace 'L3 image simulations' with 'Level-3 (L3) image simulations, i.e. resampled and optionally co-added L2 data'. Reason: L3 is not defined.
5. Replace 'on a galaxy hydro sim' with 'on a galaxy hydrodynamical simulation'. Reason: spelling out uncommon truncated words remove ambiguity and improve clarity.
6. Replace 'is pointing to (r, d) = (270, 66)' with 'is pointing to (R.A., Dec.) = (270, 66)'. Reason: r, d are not standard units.
7. Replace 'field with contains the' with 'field containing the'. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/install.html>
1. Now that we also have some experience in installing romanisim on a variety of machines, we strongly suggest to add a sentence of the very beginning with a warning saying that, first and foremost, users without FFTW C libraries MUST install them before anything else, and that failure to do so would make galsim unusable (because galsim doesn’t know where these libraries are when it is installed as a dependency of romanisim).
2. Replace 'that may cause more difficulty' with 'that may cause issues' or something along these lines. Reason: typo/grammar.
3. Replace 'presently uses supports Sersic' with 'presently supports Sersic'. Reasoning: typo/grammar.
4. 'This functionality is not available to the general community at the time of writing.' It looks like this sentence can be removed now, as CRDS is public and people should really to use --usecrds to create realistic images. In addition, --usecrds=True should be the default behavior, since otherwise galsim outputs wrong images (x-flipped).
5. The one to the last sentence starting with 'You may wish to, for example, set up a new python virtual environment' should be put near the beginning of the page, right before pip installing romanisim. Reason: some users have reported conflicts with unidentified python packages, and a clean environment appeared to fix all of them. A clean environment must be set up before any pip command is executed.
6. The last sentence can be removed as the FFTW issue affects all CPUs regardless of architecture. The new sentence suggested in (1.) should address all FFTW installation-related concerns.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/running.html>
1. Replace 'allows most simulator functionality' with 'allows most simulator functionalities'. Reason: most is always plural.
2. Please spell out what 'CLI' means.
3. Please update the 'romanisim-make-image -h' output with the one produced in the latest version of romanisim. The output currenly on the page still mentions 'fits' instead of 'asdf', for instance, and input catalogs in .csv format made with romanisim do not work.
4. Replace 'the SCA to simulate' with 'the Sensor Chip Assembly (SCA) to simulate'. Reason: SCA is not defined.
5. Replace 'this is the recommended option when CRDS is available' with 'this is the recommended option'. Reason: CRDS is available and public.
6. ‘The --webbpsf argument indicates that the WebbPSF package should be used to simulate the PSF; note that this presently disables chromatic PSF rendering.’ Later on page <https://romanisim.readthedocs.io/en/latest/api/romanisim.image.add_objects_to_image.html,> there is this statement: ‘Note: this includes Poisson noise when photon shooting is used (i.e., for chromatic source profiles), and otherwise is noise free’. As it is, it seems that there is no Poisson noise if users have the —webbpsf tag on? Please clarify.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/image.html>
1. Replace 'romanisim builds image' with 'romanisim builds images' or 'romanisim builds an image'. Reason: typo/grammar.
2. ‘for the rate of photons ‘. The sentence describes up-the-ramp fitting, so it should really be DNs not photons. Please check the unit.
3. ‘ground images see only resultants.’ Not 100% sure what this phrase means. Please add more explanation to make it clearer or remove it.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.add_objects_to_image.html>
1. ‘should be multiplied by this factor to convert to total counts in the image’. Please use ‘DNs’ instead of ‘counts’ to avoid confusion.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.gather_reference_data.html>
1. ‘This functionality is intended to allow users to specify different levels via a configuration file and not have them be overwritten by the CRDS defaults, but it’s not terribly clean.’ Can this functionality be explained in a clearer way?
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.in_bounds.html>
1. ‘keep sources up to margin outside of bounds’. A few lines below: 'whether each source lands near the image (True) or not (False)’. There appears to be contradiction (or some confusion) between these 2 statements. Are kept sources only those whose center is inside a SCA, or do they need to be “near” the SCA to be kept? and what does it mean to be near? E.g., is any part of an r_e=100 pixels bright galaxy galaxy whose center is 1 pixel outside the SCA considered in the simulation or not?
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.inject_sources_into_l2.html>
1. ‘Inject sources into an L2 image.’’, onto an existing L2 image’, ‘Source injection into an L2’. Please check that the use sof “onto” and “into” are correct.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.make_l2.html>
1. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
2. Replace ‘list of list’ with ‘list of lists’. Reason: typo/grammar.
3. ‘dark rate image to subtract from ramps (electron / s)’. Dark rate files in DRCS are in DN/s. Please clarify and fix the sentence. Also make sure that romanisim uses the correct unit for dark-rate files from CRDS.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.make_test_catalog_and_images.html>
1. ‘This routine kicks the tires on everything in this module.’ If possible, add more context and explanation to what this routine does.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.simulate.html>
1. ‘use CRDS to get distortion maps’. CRDS is used for more than just the distortion model. Please add more context.
2.Replace ’0, 1 or 2, specifying level 1 or level 2 image 0 makes a special idealized ‘counts’ image’ with ‘0, 1 or 2, specifying level 1 or level 2 image 0 makes a special idealized DN image’ to avoid confusion. Reason: L1 and L2 are in DN and DN/s and “Counts’ it too generic and confusing. Unclear what the unit the Level-0 image actually is, as there is little to no documentation about these level-0 data. We suggest to either add documentation to describe level-0 data, or not mention level 0 at all to avoid confusion.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.simulate_counts.html>
1. ‘Simulate total counts in a single SCA.’ Please replace ‘counts’ with the correct unit. Does romanisim creates these idealized scenes in e/s? If so, state it clearly.
2. ‘This gives the total counts in an idealized instrument with no systematics;’. Isn’t this a realization of the total counts taken from a Poisson distribution, as described in <https://romanisim.readthedocs.io/en/latest/romanisim/image.html?> If true, this function will never actually produce an image with the total counts of sources, but only a possible Poisson realization of them. Please clarify. Also, explicitly state the unit instead of ‘counts’.
3. Replace ‘dark rate image to use (electrons / s)’ with ‘dark rate image to use (DN/ s)’. Reason: the dark rate image in CRDS is in units of DN/s. If the dark rate is converted from DN/s to e/s via the gian, please state it in the documentation.
4. ‘do not render sources more than this many pixels off edge of detector’. Does this imply that light from sources outside the SCA can still contribute to the image? This seems to contradict text in <https://romanisim.readthedocs.io/en/latest/api/romanisim.image.in_bounds.html.> Please clarify.
5. ‘idealized image of scene as seen by Roman, giving total electron counts’, Please check that electron is the correct unit.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.simulate_counts_generic.html>
1. The first sentence ‘No Roman specific code allowed!’ is confusing.
2. ‘and the counts entering the image’. Please replace counts with the correct unit.
3. Replace ‘this role instead is played’ with ‘this role is instead played’. Reason: typo/grammar.
4. Replace ‘Then there are a few of individual components’ with ‘Then there are a few individual components’. Reason: typo/grammar.
5.’ a flat field for modulating the object and sky counts’. Please replace counts with the correct unit.
6. ‘For non-chromatic profiles, the factor converting flux to counts / s.’ Please replace counts with the correct unit.
7. ‘Image or constant with the counts / pix / sec from sky’. Please replace pix with pixel and use the correct unit instead of counts.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.trim_objlist.html>
1. ‘Trim a Table of objects’. Does ‘Table’ need to be capitalized? If so, why? Please fix or explain.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.image.trim_objlist.html>
1. ‘In contrast to inbounds, this doesn’t require the x and y coordinates of the sources, and just uses the ra/dec directly without needing to do the WCS transformation’. Does this mean that the trim is done before considering distortion? If so, hopefully the 10% extra distance is enough to account for distortion effects in all SCAs. Was this checked?
 
<https://romanisim.readthedocs.io/en/latest/romanisim/l1.html>
1. ‘the pixels count the total number of photons having entered each pixel’. Please check that the unit (photons) is correct. If the point of view is the read, it should be DN. if the point of view is a pixel, then it should be electrons, which is what pixels store. If indeed the point of view are the sources themselves, then the sentence needs to be made clearer.
2. ‘which contains the number of photons each pixel’. Please check the unit. Pixels in the detector contain electrons. Pixels in a read image contain DN.
3. Replace ‘receive absent any instrumental systematics’ with ‘receive in the absence of any instrumental systematics’. Reason: typo/grammar.
4. ‘drawing the appropriate number of photons from the total number of photons for each read following a binomial distribution’. Make sure photons is the correct unit. Pixels in the detector store electrons. Pixels in the read image are in DNs.
5. “This is slightly different than including IPC in the PSF kernel because including IPC in the PSF kernel leaves the Poisson noise uncorrelated’. The pixel value in read N does depend on the pixel value of read N-1, as reads are nondestructive, so there is still correlation regardless of where IPC is applied (or even not applied).
6. Replace ‘only applies to pixels which were’ with ‘only applies to pixels that were’. Reason: typo/grammar.
7. Replace ‘If the same pixel is receives large fluxes’ with ‘If the same pixel receives large fluxes’. Reason: typo/grammar.
8. Replace ‘If the same pixel is receives’ with ‘if the same pixel receives’. Reason: typo/grammar. On the same sentence: ‘‘If the same pixel is receives large fluxes multiple times, these are treated as two independent events’. Does this mean that multiple events are really treated as just 2 independent events? If so, why, please explain. If not, please fix the text.
9. Please fix the “sim” in the equation following ‘according to a power law distribution’.
10. ‘Convert images into L1 images, with ramps.’ Converting what images into L1 images? Please clarify.
11. Remove ‘and probably where I should start’.
12. ‘a binomial draw from the total counts weighted by the chance that each count landed in this particular time window.’ Please replace “counts’ with the correct unit.
13. Please fix the reference in ‘Following Casertano (2022?)’. It is indeed 2022 <https://www.stsci.edu/files/live/sites/www/files/home/roman/_documents/Roman-STScI-000394_DeterminingTheBestFittingSlope.pdf>
14. Replace ‘So that gives the variance from Poisson counts in resultant’ with ‘So that gives the variance from Poisson counts in resultants’ or ‘So that gives the variance from Poisson counts in a resultant’. Reason: typo/grammar
15. ‘I can separately control the mean’. This party of the documentation seems like a copy and paste from a personal notebook. Please rephrase. There are more instances like this in the page.
16. Replace ‘The poisson noise is a sum of scaled’ with ‘The Poisson noise is a sum of scaled’. Reason: typo/grammar.
17. All in all, the entire description of the romanisim.l1.module should be rewritten. As it is now, it looks like a list of personal l notes: “oh, maybe I can do that” or “oh, what if I do this other thing” rather than actual documentation. The text should simply describe what the module does.
18. Replace “The detector is accumulating photons’ with ‘The detector is accumulating electrons’.
19. Replace ‘so the mapping between accumulated photons’ with ‘so the mapping between accumulated electrons’.
20. ‘When simulating an image, the photons entering each pixel are simulated’. Is this true? Aren’t electrons simulated instead? Please double check.
21. ‘with this mapping into observed counts’. Replace counts with the correct unit.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.add_ipc.html>
1. Replace ‘resultants describing scene’ with ‘resultants describing a scene’. Reason: typo/grammar.
2. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.add_read_noise_to_resultants.html>
1. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
2. Replace ‘list of list’ with ‘list of lists’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.apportion_counts_to_resultants.html>
1. Replace ‘appropriate assignment of counts to each read composing a resultant’ with ‘appropriate assignment of DNs to each read composing a resultant’. Reason: DN is the correct unit here.
2. ‘On some systems I’ve used Poisson draws’. Later on: ‘We loop’, ‘We then’, ‘We accumulate’. We suggest to rephrase these sentences to make them look like a documentation.
3. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
4. Replace ‘list of list’ with ‘list of lists’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.make_asdf.html>
1. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.make_l1.html>
1. Replace ‘This apportions the total counts ‘ with ‘This apportions the total DNs ‘. Reason: DN is the correct unit here.
2. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.read_pattern_to_tij.html>
1. Replace ‘list of list’ with ‘list of lists’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.tij_to_pij.html>
1. Replace ‘we scale these probabilities’ with ‘these probabilities are scales’. Reason: avoid using “I” and “we”.
2. Replace ‘list of list’ with ‘list of lists’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.l1.validate_times.html>
1. Replace ‘Verify that a set of times tij for a valid resultant’ with ‘Verify that a set of times tij is ascending for a valid resultant’. Reason: missing verb in the subordinate clause.
2. Replace ‘list of list’ with ‘list of lists’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.nonlinearity.evaluate_nl_polynomial.html>
1. Replace ‘the observed counts for non-linearity’ with ‘the observed DNs for non-linearity’. Reason: DN is the correct unit.
2. Replace ‘As photons arrive, they make it harder for the device to count future photons due to classic’ with ‘As electrons are accumulated into pixels, they make it harder for the device to accumulate more electrons due to classical’. Reason: it is the number of electrons already stored in a pixel and not the arrival of photons that affect linearity.
3. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
4. ‘This function converts some observed counts…’. Please clarify what is meant here with counts. The rest of the sentence is also not easy to to understand. Please rephrase to improve clarity.
5. ‘Number of counts already in pixel’. Please replace counts with the correct unit.
6. ‘The corrected number of counts’. Please replace counts with the correct unit.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.nonlinearity.repair_coefficients.html>
1. Replace ‘with no-op coefficients’ with ‘with identity-transformation coefficients’. Reason: no-op is not the correct adjective in this context.
2. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.nonlinearity.NL.html>
1. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
2. Replace ‘Gain (electrons / count) for converting counts to electrons’ with ‘Gain (electron / DN) for converting DNs to electrons’. Reason: DN is the correct unit.
3. ‘Compute the correction of observed to true counts’. Please replace counts with the correct unit, or specify that this function can correct both electrons and DNs (if this is the case).
4. Replace ‘Accrdingly’ with ‘Accordingly,’. Reason: typo/grammar.
5. ‘¯ The corrected counts.’ Remove the ‘-‘ and the bold face. Replace counts with the correct unit, or specify that this function can correct both electrons and DNs (if this is the case).
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.persistence.fermi.html>
1. Replace ‘For influence level below’ with ‘For fluence level below’. Reason: typo/grammar.
2. Fix the format of the ‘Returns:’ text.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.cr.create_sampler.html>
1. Replace ‘ tabulated values which can be’ with ‘ tabulated values that can be’. Reason: typo/grammar.
2. ‘1-d’. All documentation pages so far make use of 1D (or 2D) instead of 1-d (or 2-d).  Please use 1D for consistency.
3. Replace ‘distribution function which allows’ with ‘distribution function that allows’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.cr.moyal_distribution.html>
1.  Replace ’1-d’ with ‘1D’ for consistency.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.cr.power_law_distribution.html>
1.  Replace ’1-d’ with ‘1D’ for consistency.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.cr.simulate_crs.html>
1. Replace ‘2-d’ with ‘2D’ for consistency.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.cr.traverse.html>
1. Please replace all instances of ‘pix’ with ‘pixels’. Reason: pix is not a standard unit.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/l2.html>
1. Replace ‘how counts get apportioned’ with ‘how DNs get apportioned’. Reason: DN is the correct unit.
2. ‘because the covariance matrices for each pixel for a one-dimensional family depending only on the ratio of the flux in the pixel to the read variance’. It looks like a verb a missing, or maybe ‘depending’ should be replaced with ‘depend’(?).
3. Replace ‘The second approach follows Casertano+2022’ with ‘The second approach follows Casertano (2022)’. Reason: that paper has a single author.
4. ‘These are then interpolated between for each pixel’. This sentence is confusing. Maybe ‘between’ can be removed(?).
5. Throughout the page, there are a lot of philosophical arguments about what should and should not be done when fitting ramps, and what should work best for what case. All of this can simply be removed in favor of just describing what the module actually does.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.ramp.fit_ramps_casertano.html>
1. Replace ‘following Casertano+2022’ with ‘following Casertano (2022)’. Reason: that paper has a single author.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.ramp.fit_ramps_casertano_no_dq.html>
1. Replace ‘following Casertano+2022’ with ‘following Casertano (2022)’. Reason: that paper has a single author.
2. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.ramp.ki_and_variance_grid.html>
1. Replace ‘needed to do ramp fitting form essentially’ with ‘needed to do ramp fitting essentially form’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.ramp.resultants_to_differences.html>
1. Replace ‘This is essentially’ with ‘This essentially is’. Reason: typo/grammar.
2. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.ramp.RampFitInterpolator.html>
1. Replace ‘we can up front calculate them overa grid in the flux’ with ‘they can be calculated up front over a grid in the flux’. Reason: typo/grammar and remove “we”.
2. Replace ‘for each of a 16 million pixels’ with ‘for each of the 16 million pixels’. Reason: typo/grammar.
3. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/romanisim/refs.html>
1. Add linearity to the bullet list of reference files used from CRDS.
2. The dark files in CRDS already contain a 2D array of the dark rate image in units of DN / s. Why is a scaling of the last resultant by its effective exposure time used, instead?
3. Replace ‘the number of photons each pixel would record’ with ‘the number of electrons each pixel would record’. Reason: pixels record electrons.
4. Please remove or rephrase the sentence: ‘I haven’t actually seen any such values yet and so haven’t tried to address that case (e.g., by clipping them).’
5. Make sure the correct unit is used in ‘Photons from the idealized “counts” image’.
6. Replace “down to ADU” with “down to DNs”.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.catalog.image_table_to_catalog.html>
1. Replace ‘Read a astropy Table’ with ‘Read an astropy Table’. Reason: typo/grammar.
2. Please spell out ‘hydro sim’.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.catalog.make_galaxies.html>
1. ‘generate sources Note this magnitude’. It looks like a full stop is missing.
2. Replace ‘a “fiducial” band which is not observed’ with either ‘a “fiducial” band, which is not observed’ or ‘a “fiducial” band that is not observed’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.catalog.make_image_catalog.html>
1. Replace ‘hydro’ with ‘hydrodynamical’. Reason: improve clarity.
2. Fix format in the ‘Returns:’ text.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.catalog.make_stars.html>
1. ‘generate sources Note this magnitude’. It looks like a full stop is missing.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.catalog.table_to_catalog.html>
1. Replace ‘Read a astropy Table’ with ‘Read an astropy Table’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/image-input.html>
1. Replace ‘hydro’ with ‘hydrodynamical’. Reason: improve clarity.
2. ‘specifies the id’. Shouldn’t this be ‘ID’, as in identification?
 
<https://romanisim.readthedocs.io/en/latest/romanisim/bandpass.html>
1. Replace 'because catalog fluxes are specified in “maggies” (i.e., in linear fluxes on the AB scale)’ with ‘because catalog fluxes are specified in “maggies” (i.e., in fluxes computed as 10^(-0.4*AB mag)’. Reason: fluxes are naturally linear, so saying ‘linear fluxes’ might be confusing, and it makes little sense to say that fluxes are on the AB scale.
2. There are a lot of instances of ‘photon(s)’ in this page. Please check that all these instances are correct. Typically, photometric zero points concern transformations from electrons (or electron/s, or, sometimes, DNs or DN/s) to magnitudes or other physical units.
3. ‘The primary purpose of this module is to provide the number of counts per second’. Please replace ‘counts’ with the appropriate unit.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.bandpass.compute_abflux.html>
1. ‘How many photons would a zeroth magnitude AB star deposit in Roman’s detectors in a second?’ Do you mean how many electrons per second are generated in an SCA by a 0th AB mag star? Please fix or improve clarity.
2. ‘lookup table of zero point fluxes for each filter (photons / s)’. I really don’t think the unit is correct. Please check.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.bandpass.compute_count_rate.html>
1. ‘How many photons would an object with SED given by flux deposit in Roman’s detectors in a second?’. Please use the correct unit. Should it be ‘How many electrons … are generated in …?’
2. ‘the total bandpass flux (photons / s)’. Please replace photons/s with the correct unit.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.bandpass.get_abflux.html>
1. ‘the zero point flux (photons / s)’. Please replace photons/s with the correct unit.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/psf.html>
1. Replace ‘implementation of the Roman point spread function’ with ‘implementation of the Roman point spread function (PSF)’. Reason: define PSF and improve clarity.
2. Replace ‘FFTs are used to to do’ with ‘FFTs are used to do’. Reason: typo/grammar.
3. The bullet point starting with ‘how do we want …’ should be rephrased or potentially removed.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.psf.make_psf.html>
1. Replace ‘via interpolation between the four corners’ with ‘via interpolation between the four corners of an SCA’. Reason: improve clarity.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/wcs.html>
1. ‘galsim.roman has an implementation of Roman’s WCS based on some SIP coefficients for each SCA. This is presumably plenty good’. It might be plenty good for some applications, but not for others. This remark should be removed as it is not universally true. In addition, galsim appears to produce wrong output images (x-flipped) when CRDS is not used.
2. Replace ‘for a given sca’ with ‘for a given SCA’. Reason: typo/grammar. This happens twice.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.wcs.fill_in_parameters.html>
1. ‘Metadata dictionary Dictionaries’. Is this correct? Why is the second Dictionaries capitalized?
2. ‘world coordinates at V2 / V3 ref (boresight or center of WFI CCDs)’. The WFI does not have CCDs. Please replace ‘CCDs’ with detectors.
3. ‘position angle (North to YIdl)’. Please describe what YIdl mean.
4. Replace ‘or the center of the WFI CCD array’ with ‘or the center of the WFI layout’. Reason: there are no CCDs in the WFI.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.wcs.get_mosaic_wcs.html>
1. Replace ‘for a given sca’ with ‘for a given SCA’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.wcs.get_wcs.html>
1. Replace ‘for a given sca’ with ‘for a given SCA’. Reason: typo/grammar.
2. Replace ‘If True, use crds reference’ with ‘If True, use CRDS reference’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.wcs.make_wcs.html>
1. Replace ‘The v2 coordinate’ with ‘The V2 coordinate’. Reason: typo/grammar.
2. Replace ‘The v3 coordinate’ with ‘The V3 coordinate’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/parameters.html>
1. ‘the default read noise when CRDS is not used’. Please state what the default value is.
2. ‘the number of border pixels’. Is this about reference pixels? If so, replace ‘border’ with ‘reference’. Also, is it just the number of them in the parameter module, or also their geometry?
3. ‘the read time’. Please state what the default value is.
4. ‘the default saturation limit’. Please state what the default value is.
5. ‘The default IPC kernel’. Please state what the default kernel values are.
6. Replace ’to which to place the given ra/dec’ with ‘o which to place the given R.A./Dec.’. Reason: typo/grammar.
7. ’the default cosmic ray parameters’. Please state what the default parameters are.
8. ‘the default gain when CRDS is not used’. Please state what the default value is.
 
<https://romanisim.readthedocs.io/en/latest/romanisim/util.html>
1. Replace ‘Compute the King (1962) profile’ with ‘Compute a King profile’. Reason: the King profile is as common as a Sersic profile and a reference is not needed.
2. Replace ‘Sample distances from a King (1962) profile’ with ‘Sample distances from a King profile’. Reason: same as above.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.celestialcoord.html>
1. Replace ‘to transform into an galsim’ with ‘to transform into a galsim’. Reason: typo/grammar.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.decode_context_times.html>
1. ‘numpy.ndarray’ is in italic for some reason.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.default_image_meta.html>
1. Fix the format in the ‘Returns:’ text (currently bold face).
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.king_profile.html>
1. Replace ‘Compute the King (1962) profile’ with ‘Compute a King profile’.
2. Fix the format in the ‘Returns:’ text (currently bold face).
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.random_points_at_radii.html>
1. Fix the format in the ‘Returns:’ text (currently bold face).
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.random_points_in_cap.html>
1. Fix the format in the ‘Returns:’ text (currently bold face).
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.random_points_in_king.html>
1. Fix the format in the ‘Returns:’ text (currently bold face).
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.sample_king_distances.html>
1. Replace ‘from a King (1962) profile’ with ‘from a King profile’.
2. Replace ‘Distances distributed according to a King (1962) profile’ with ‘Distances distributed according to a King profile’.
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.scalergb.html>
1. Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].
 
<https://romanisim.readthedocs.io/en/latest/api/romanisim.util.update_aperture_and_wcsinfo_metadata.html>
1. Replace ‘to the v2 and v3 of the center’ with ‘to the V2 and V3 of the center’. Reason: typo/grammar.

schlafly:
Closes #152.  

I adopted most of the changes wholesale.  Here is a list of some of the ones where more conversation might be beneficial:

* Overall, there is quite a bit of confusion throughout the documentation related to units. There is about a factor of 2 difference between electrons in the pixels/detectors and DNs in the read arrays, but both are usually referred to as ‘counts, which is a very generic term. Also, sometimes the word ‘photon(s)’ seems to be used instead of ‘electron(s)’, which make readers think the sentence is not describing the detector but the source. Please replace relevant instances of “photon(s)” and “count(s)” with the correct units (either electron(s) or DN(s)) according to the context. This will help clarify a lot of confusing statements.

We adopted electrons when referring to Poisson-distributed quantities, and DN when appropriate.
 
* Also, there are a few pages, mostly related to the description of L1 and L2 images, containing what appears to be a list of personal notes and thoughts rather than a description/documentation of what the relevant functions/methods do. We strongly suggest to rewrite those pages in the spirit of a documentation style, and avoid the use of “I” or “we” subjects as much as possible.

This was trimmed slightly and prefaced slightly, but I find some of that discussion useful and did not feel the need to just delete it.
 
* The “Search docs” bar does not work. This might not be a romanisim-only issue, as the issue appears to be spread out to most/all readthedocs pages, but it should still be fixed.

This looks to be fixed with a config change.
 
* In some pages equations are given using a latex interpreter while in others are simply typed down in ASCII, which are less easy to read. We suggest to always use the latex interpreter for clarity.

We updated these when called out in the detailed notes.
 
* Sometimes GalSim is rendered as ‘GalSim’, other as ‘galsim’ (even in the same page, e.g., https://romanisim.readthedocs.io/en/latest/romanisim/wcs.html). Also, WebbPSF is always rendered as ‘webbpsf’. Can these packages be consistently rendered throughout the documentation? If a distinction is made to distinguish between GalSim as the name of the program, and galsim as the python package, can a different font be used for python packages, to improve clarity?

I haven't changed the styling here.

* 'This functionality is not available to the general community at the time of writing.' It looks like this sentence can be removed now, as CRDS is public and people should really to use --usecrds to create realistic images. In addition, --usecrds=True should be the default behavior, since otherwise galsim outputs wrong images (x-flipped).

I've deleted the line about CRDS not being public.  I have not changed the default; the present default matches GalSim's default, which is e.g. what was used for the large PIT simulations, and CRDS has its own issues (e.g., the linearity reference files are of marginal quality).

* The one to the last sentence starting with 'You may wish to, for example, set up a new python virtual environment' should be put near the beginning of the page, right before pip installing romanisim. Reason: some users have reported conflicts with unidentified python packages, and a clean environment appeared to fix all of them. A clean environment must be set up before any pip command is executed.

This is general python stuff and has no particular relationship to romanisim; I have not made a change.

https://romanisim.readthedocs.io/en/latest/romanisim/running.html
Please update the 'romanisim-make-image -h' output with the one produced in the latest version of romanisim. The output currenly on the page still mentions 'fits' instead of 'asdf', for instance, and input catalogs in .csv format made with romanisim do not work.

* ‘The --webbpsf argument indicates that the WebbPSF package should be used to simulate the PSF; note that this presently disables chromatic PSF rendering.’ Later on page https://romanisim.readthedocs.io/en/latest/api/romanisim.image.add_objects_to_image.html, there is this statement: ‘Note: this includes Poisson noise when photon shooting is used (i.e., for chromatic source profiles), and otherwise is noise free’. As it is, it seems that there is no Poisson noise if users have the —webbpsf tag on? Please clarify.

add_objects_to_image does not add Poisson noise if --chromatic is false, and so that nois emust be added at other stages of the simulation.  We do this, e.g., in L2, L3, L2 injection, and L3 injection.  I have not changed the text.

* ‘for the rate of photons ‘. The sentence describes up-the-ramp fitting, so it should really be DNs not photons. Please check the unit.

I haven't changed this.  We convert the DN to electrons and then do the fit, in the same way as the pipeline.  The result is then converted back to DN / s for the final L2 images using the gain.

* ‘dark rate image to subtract from ramps (electron / s)’. Dark rate files in DRCS are in DN/s. Please clarify and fix the sentence. Also make sure that romanisim uses the correct unit for dark-rate files from CRDS.

I haven't changed this.  romanisim converts the dark from DN / s to e / s before sending it to this routine.
 
* 2.Replace ’0, 1 or 2, specifying level 1 or level 2 image 0 makes a special idealized ‘counts’ image’ with ‘0, 1 or 2, specifying level 1 or level 2 image 0 makes a special idealized DN image’ to avoid confusion. Reason: L1 and L2 are in DN and DN/s and “Counts’ it too generic and confusing. Unclear what the unit the Level-0 image actually is, as there is little to no documentation about these level-0 data. We suggest to either add documentation to describe level-0 data, or not mention level 0 at all to avoid confusion.

I've added that the level 0 images are only intended for testing purposes, and changed counts to "total electron".

* ‘This gives the total counts in an idealized instrument with no systematics;’. Isn’t this a realization of the total counts taken from a Poisson distribution, as described in https://romanisim.readthedocs.io/en/latest/romanisim/image.html? If true, this function will never actually produce an image with the total counts of sources, but only a possible Poisson realization of them. Please clarify. Also, explicitly state the unit instead of ‘counts’.

I have changed total counts to total electrons.  I have added that this includes Poisson noise.

* Replace ‘dark rate image to use (electrons / s)’ with ‘dark rate image to use (DN/ s)’. Reason: the dark rate image in CRDS is in units of DN/s. If the dark rate is converted from DN/s to e/s via the gian, please state it in the documentation.

This input takes the dark rate image in electrons / s.  The conversion from the native CRDS units to the dark rate is done elsewhere in the simulation.

* ‘do not render sources more than this many pixels off edge of detector’. Does this imply that light from sources outside the SCA can still contribute to the image? This seems to contradict text in https://romanisim.readthedocs.io/en/latest/api/romanisim.image.in_bounds.html. Please clarify.

This is exactly the argument "margin" passed to in_bounds.  The docstring for in_bounds has been modestly updated.

* ‘In contrast to inbounds, this doesn’t require the x and y coordinates of the sources, and just uses the ra/dec directly without needing to do the WCS transformation’. Does this mean that the trim is done before considering distortion? If so, hopefully the 10% extra distance is enough to account for distortion effects in all SCAs. Was this checked?

I've updated this text.  The point is that this doesn't get each source's x & y coordinates, and instead just gets everything in a circle with a radius of 1.1 times the center-corner distance.  The center-corner distance uses the full WCS and distortion.  I have not checked that this is always adequate, but I would be surprised if it were not.  Let me know if you find otherwise.

* ‘the pixels count the total number of photons having entered each pixel’. Please check that the unit (photons) is correct. If the point of view is the read, it should be DN. if the point of view is a pixel, then it should be electrons, which is what pixels store. If indeed the point of view are the sources themselves, then the sentence needs to be made clearer.

I've changed to electrons.

* ‘which contains the number of photons each pixel’. Please check the unit. Pixels in the detector contain electrons. Pixels in a read image contain DN.

I've changed to elecetrons.

* ‘drawing the appropriate number of photons from the total number of photons for each read following a binomial distribution’. Make sure photons is the correct unit. Pixels in the detector store electrons. Pixels in the read image are in DNs.

Changed to electrons.

* “This is slightly different than including IPC in the PSF kernel because including IPC in the PSF kernel leaves the Poisson noise uncorrelated’. The pixel value in read N does depend on the pixel value of read N-1, as reads are nondestructive, so there is still correlation regardless of where IPC is applied (or even not applied).

I added "between pixels."  Of course the correlation between reads and resultants doesn't depend on IPC, but IPC correlates one pixel with its neighbors.

* Replace ‘If the same pixel is receives’ with ‘if the same pixel receives’. Reason: typo/grammar. On the same sentence: ‘‘If the same pixel is receives large fluxes multiple times, these are treated as two independent events’. Does this mean that multiple events are really treated as just 2 independent events? If so, why, please explain. If not, please fix the text.

The point here is that in past exposures the same pixel can be hit by a bright source multiple times.  e.g., a bright star in the previous exposure and the two-previous exposure.  Sanchez+2023 doesn't have a particular model for what the persistence flux is as a function of time for that case.  In the simulation we model this as the persistence flux that would be left over from the exposure two exposures ago plus the persistence flux that would be left over from the exposure one exposure ago.  I have slightly tweaked the text.

* Replace ‘appropriate assignment of counts to each read composing a resultant’ with ‘appropriate assignment of DNs to each read composing a resultant’. Reason: DN is the correct unit here.

Changed to electrons.  This phase of the simulation is working with electrons.

* Replace ‘This apportions the total counts ‘ with ‘This apportions the total DNs ‘. Reason: DN is the correct unit here.

Replaced with electrons, because this is what is done.

* Throughout the page, there are a lot of philosophical arguments about what should and should not be done when fitting ramps, and what should work best for what case. All of this can simply be removed in favor of just describing what the module actually does.

I have left in most of this discussion.

* The dark files in CRDS already contain a 2D array of the dark rate image in units of DN / s. Why is a scaling of the last resultant by its effective exposure time used, instead?

This was out of date and has been replaced; the documentation was from before the dark rate column existed.

* ‘galsim.roman has an implementation of Roman’s WCS based on some SIP coefficients for each SCA. This is presumably plenty good’. It might be plenty good for some applications, but not for others. This remark should be removed as it is not universally true. In addition, galsim appears to produce wrong output images (x-flipped) when CRDS is not used.

Changed to "could presumably be adequate."

* https://romanisim.readthedocs.io/en/latest/romanisim/parameters.html

I did not add the default values for all of these quantities, since I imagine they may change.  I did add a link to the code where they are all defined.  In some cases where I found the values very unlikely to change I added specific values.

* ‘the number of border pixels’. Is this about reference pixels? If so, replace ‘border’ with ‘reference’. Also, is it just the number of them in the parameter module, or also their geometry?

I changed this to the number of border reference pixels (4).

* https://romanisim.readthedocs.io/en/latest/api/romanisim.util.scalergb.html
* Please fix the dimension order of numpy arrays: F[z,y,x] not F[z,x,y]. Similar for F[y,x] not F[x,y].

Here I've changed the docstring to be np.ndarray[ny, nx, 3] from np.ndarray[npix, npix, 3].  I don't think it's a good idea to describe the last axis, which has the three color channels, as the x axis.


schlafly:
@AndreaBellini, let me know if you're interested in looking over these doc changes before I merge.

schlafly:
Andrea signed off out of band, so I'm going ahead and merging this.

PR #162, opened 2024-10-03
RCAL-853: Allow ability to introduct velocity aberration scale factor

stscieisenhamer:
WIP [RCAL-853](https://jira.stsci.edu/browse/RCAL-853)

Main PR: [PR#1354](https://github.com/spacetelescope/romancal/pull/1354/files)

This PR adds the velocity aberration transformation to the WCS to determine the FOV of the resulting L1 simulations. In addition, the meta `velocity_aberration.scale_factor` is populated so that the ELP pipeline step `assign_wcs` correctly accounts for the applied aberration.


schlafly:
Jonathan and I talked about this on Friday, but recording some of my thoughts here.

I think we need to add some bits around here https://github.com/spacetelescope/romanisim/blob/main/romanisim/util.py#L302-L309 .  The issue is that this correction produces the right WCS (i.e., the one that would be observed if the aberrated WFI_CEN were at the reference ra/dec.  But there is a bit of code to populate the corresponding wcsinfo centered at the detector, and I don't think we'll do the right thing there at present.  In particular that code assumes that wcs(center) = radec_ref, but I think for the romancal WCS that's not quite right when aberration is non-zero.  Instead the point whose v2v3a = center goes to radec_ref.  Probably the easiest way to get that point is to extract the part of the WCS from v2v3a to sky and evaluate radec_ref = v2v3a_to_sky(v2v3).  Likewise roll_ref needs to work from this frame and so radecn need to be evaluated there.

I think that's everything?  But the proof at the end of the day is a check that the romanisim WCS exactly matches the romancal WCS for some non-trivial scale factor.

We also should populate the ephemerides so that the scale factors get computed.  I suggest doing this by assuming that the telescope is exactly at L2 and neglecting more details of the orbit.

romanisim/velocity_aberration.py
```diff
@@ -0,0 +1,93 @@
...
+def compute_va_effects(velocity_x, velocity_y, velocity_z, ra, dec):
+    """ Computes constant scale factor due to velocity aberration as well as
+    corrected ``RA`` and ``DEC`` values.
+
+    Parameters
+    ----------
+    velocity_x, velocity_y, velocity_z: float
+        The components of the velocity. These are celestial coordinates, with x toward the
+        vernal equinox, y toward right ascension 90 degrees and declination
+        0, z toward the north celestial pole.
```
schlafly, line 68: ICRS, presumably?

romanisim/util.py
```diff
@@ -539,3 +561,32 @@ def merge_dicts(a, b):
...
+        point, such as the center of a detector) in the barycentric coordinate
+        system.  The equator and equinox should be the same as the coordinate
+        system for the velocity. In degrees.
+
+    Returns
+    -------
+    scale_factor : float
+        The velocity aberration scale factor
+    """
+    _, velocity = get_body_barycentric_posvel('earth', date)
```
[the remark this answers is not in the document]
schlafly: FWIW, I was worried that this was going to be in some solar-system focused frame, but it's not, it's in ICRS barycentric coordinates and is exactly what we want.

romanisim/velocity_aberration.py
```diff
@@ -0,0 +1,93 @@
...
+
+    u_beta = np.dot(u, beta)
+    igamma = np.sqrt(1.0 - beta2)  # inverse of usual gamma
+    scale_factor = (1.0 + u_beta) / igamma
+
+    # Algorithm below is from Colin Cox notebook.
+    # Also see: Instrument Science Report OSG-CAL-97-06 by Colin Cox (1997).
+    u_corr = (igamma * u + beta * (1.0 + (1.0 - igamma) * u_beta / beta2)) / (1.0 + u_beta)
+
+    return scale_factor, u_corr
```
schlafly, line 56: Have you tested one comparison with Webb?  We should get close to the same scale factor that they get for some random observation.

schlafly (commented):
This all looks good to me.  Can I ask you to add one of your consistency check plots to this PR, and one or two paragraphs to the docs here?

https://romanisim.readthedocs.io/en/latest/romanisim/wcs.html

Please mention the typical scale of the correction (i.e., a value of va_scale).  I'd also mention in a sentence that we make a wcsinfo for each individual detector but internally our WCS in centered on wfi_cen or the boresight, and differing treatment of the spherical to cartesian transformations makes a ~10^-6 arcsec effect.

romanisim/velocity_aberration.py
stscieisenhamer: For the group stats I had done with nircam, the scale factor ranges from 0.9999020030807361 to 1.0000974957714621.  The range of the romanisim testing hast been from 1.0 to 1.000055805739964.

schlafly: I mean, if we simulate an image at the same time of year as Webb, or with the same velocity, do you get the same answer?

stscieisenhamer:
Docs updated. Also changed the command-line default to use a scale of 1.0.

romanisim/velocity_aberration.py
```diff
@@ -0,0 +1,93 @@
+"""Calculate velocity aberration based on velocities"""
+
+import logging
+import numpy as np
+from gwcs.geometry import SphericalToCartesian, CartesianToSpherical
+from scipy.constants import speed_of_light
+
+# Configure logging
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
```
schlafly, line 10: Just since I see it... the policy in other places in romanisim has been to do something like
```
from romanisim import log
log.info(...)
```
This has never worked great, which I have blamed on weird interactions with stpipe.  Is this doing something good on top of that?

stscieisenhamer: The form in the code is generally used for library/utility modules. This basically leaves the logging up to whatever higher level code is using the module. For edification, [this is where it comes from](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

However, if the rest of romanisim uses the another standard, as suggested, there is no real issue in changing it. I'll change it so it is consistent with the rest of the package.

About the weirdness with `stpipe`: The is unfortunately an `stpipe` problem, mainly because `stpipe` violates what is suggested in the aforementioned link. Without some hoop-jumping, until `stpipe` gets it *stuff* together, there is not much to do about the weirdness.

schlafly: Right.  But I think that that code would belong around here?
https://github.com/spacetelescope/romanisim/blob/main/romanisim/__init__.py#L15-L20
Then other modules wouldn't need it?  I might be missing something about what doing it in romanisim.velocity_aberration does relative to doing things module-wide.

And it looks like I don't set a default handler there, and maybe I should?  Then all of the other modules would likewise benefit.

I will certainly admit that to me it seems python backed itself pretty far into a corner on logging, which is hard to get right!

stscieisenhamer:
Do not understand the build issues, but those will be unrelated to this PR. Merging away.

PR #163, opened 2024-10-03
Add Dependabot to update github actions and pin actions to hash

braingram:
This PR pins github actions to hashes (rather than versions) as recommended by github:
https://wellarchitected.github.com/library/application-security/scenarios-and-recommendations/actions-security/#pin-versions-of-actions
and enables dependabot to check these monthly. Dependabot will open PRs requesting updates when new versions of actions are released.


braingram:
I can't request reviewers. @zacharyburnett and @schlafly would you take a look? Thanks!

PaulHuwe (approved):
LGTM

braingram:
Thanks! I have no green "merge" button to push so feel free to merge this when convenient.

PR #165, opened 2024-10-08
Remove units from roman data model outputs

schlafly:
This PR removes the units from output data models, to reflect changes in rad / roman_datamodels that remove units there.

It also includes two extraneous doc fixes (sorry!)


schlafly:
@zacharyburnett , does this MacOS python 3.13 failure make sense to you?  We once barred python >= 3.13, but somehow the MacOS job wants to use it and then fails.

zacharyburnett:
Yeah, looks like when unpinned, Tox grabs the latest Python without checking if the project can support it, and the install fails because `romanisim` explicitly doesn't want 3.13; we should test unpinning Python, and if that doesn't work change the toxenv in the CI to `py312-xdist`

PR #166, opened 2024-10-18
Update romanisim to use the new L1 / L2 metadata

schlafly:
We are updating the metadata in rad/rdm for L1 and L2 files.  This updates romanisim to use that new metadata format.  It also changes the file names to use wfi in lowercase rather than uppercase.


schlafly:
@PaulHuwe , I think this is basically working; tests are currently failing because BORESIGHT / WFI_CEN / CGI_CEN aren't allowed aperture names, but I think we want to add those to the allowed list.  When the new rad/rdm PRs go in we'll want to use this PR to make the L1 data that can be used for the romancal regression tests.

schlafly:
Okay, this is now ready to go when we merge the rad/rdm changes.

PR #167, opened 2024-10-21
remove upper pin from `requires-python`

zacharyburnett:


zacharyburnett:
It looks like we're still in a limbo state, where the Linux runner grabs Python 3.12 as "latest 3" but the macOS runner grabs Python 3.13

schlafly:
Tests passing, so I guess you can go ahead and un-draft it and I'll merge?  Thanks!

PR #172, opened 2024-11-12
Change CRDS keyword names.

schlafly:
This updates the names of the keywords used to specify the CRDS software version and context used in romanisim.


schlafly:
Tests are passing and this is pretty simple; going ahead and merging.

PR #176, opened 2024-11-20
allow data workflows (WebbPSF + GALSIM) to be used by upstream repos for downstream testing

zacharyburnett:
we're attempting to test `romanisim` as part of `gwcs`'s downstream testing, and it would be very useful to be able to use this workflow to set up the webbpsf + GALSIM data cache for there:
- https://github.com/spacetelescope/gwcs/pull/530

This involves adding the `workflow_call` trigger and also allowing the user to specify the download path for the data


schlafly (approved):
Thanks, looks good to me, and tests are passing.

PR #177, opened 2024-11-20
Updating the MA tables in parameters.py and setting the default MAtable to 4 in ris_make_utils.py

eunkyuh:
- Updating the MA tables within parameters.py to match PRD15. Excluded 3 MA tables: 1(DEFOCUS_MOD), 2(DEFOCUS_LRG), and 18(DIAGNOSTIC)

- Updating rist_make_utils.py to change the default MA table to 4 as the previous selection of MA table #1 is not available anymore.


schlafly (approved):
Looks good to me, thanks!

PR #190, opened 2024-12-16
Drop unused dependencies

braingram:
An attempted regtest run with python 3.13 showed a failure when the test machine failed to build `getch`:
https://github.com/spacetelescope/RegressionTests/actions/runs/12264011982/job/34217036983?pr=183#step:14:38
I don't see any uses of this package. (FWIW The python 3.13 tests aren't failing in the CI here so the above failure could be seen as a bug in the proposed regtest test machine changes).

This PR removes unused dependencies:
- getch: looks like a dead package (no release since 2013): https://pypi.org/project/getch/
- colorama: this isn't imported
- flake8: this was replaced by ruff

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,6 @@ docs = [
 test = [
     "ci-watson >=0.3.0",
     "colorama >=0.4.1",
-    "getch >=1.0.0",
     "pytest >=4.6.0, <=8.0",
     "pytest-openfiles >=0.5.0",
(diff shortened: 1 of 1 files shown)


schlafly:
Thanks!

schlafly:
(and FWIW, e.g. getch and colorama date back to the original skeleton commits before I started and we actually had written any non-boilerplate code, so ... no problem dropping them!)

PR #197, opened 2025-01-24
Update for unit-free CRDS reference files.

schlafly:
This PR updates the reference data gathering routine to accommodate the new, unit-free reference files.  It achieves this by adding the units back after the files have been loaded.  This is a little goofy as those units are then again later stripped and so we could consider a more thorough-going removal of units from romanisim, but it works.

It also adds "crds://" to the file names of crds reference files.

It includes two unrelated changes:
1. It updates the documentation to be more explicit about the numerator and denominator of 'ba' in the catalogs.
2. It renames the simulated catalog extension of the L3 files to have the same name as the one for the L2 files.


PaulHuwe (approved):
This is okay as a temporary fix, but obviously brittle. This further emphasizes the need for unit streamlining (and commenting) throughout the code.

pyproject.toml
```diff
@@ -14,19 +14,19 @@ classifiers = [
...
     "defusedxml >=0.5.0",
     "galsim >=2.5.1",
-    "rad >=0.22.0",
-    "roman_datamodels >=0.22.0",
+    "rad >=0.23.0",
+    "roman_datamodels >=0.23.0",
     "gwcs >=0.19.0",
     "jsonschema >=4.8",
-    "numpy >=1.22",
+    "numpy >=1.24",
```
nden, line 29: not sure this is the cause but change line 5 for the required python  to be ">=3.11"

schlafly:
Thanks both, sorry I had missed this!

PR #199, opened 2025-02-05
RCAL-959 COSMOS and GAIA Sources in Catalogs

PaulHuwe:
Adding methods to select random COSMOS galaxies and/or GAIA stars to make more realistic catalogs.


PaulHuwe:
Sample Level 3 Mosaic with radius = 0.1 degree (can see the area search):
RSIM_cosmos_gaia_imviz-0

PaulHuwe:
Histogram of stellar and galaxy densities (wrt. all objects):
newplot (21)

Same histogram of stellar and galaxy densities, but wrt. their individual classes:
newplot (22)

PaulHuwe:
Histogram of rsim's selected cosmos sources vs. the full cosmos catalog (densities wrt. each dataset):
newplot (16)

PaulHuwe:
Setting this ready for review, as the failing tests are in main, not this PR.

romanisim/catalog.py
```diff
@@ -241,6 +435,76 @@ def make_galaxies(coord,
...
+        bandpasses = ["F062", "F087", "F106", "F129", "F146", "F158", "F184", "F213"]
+
+    if date is None:
+        date = Time('2026-01-01T00:00:00')
+
+    if rng is None:
+        rng = galsim.UniformDeviate(seed)
+
+    if rng is None:
+        rng = galsim.UniformDeviate(seed)
```
schlafly, line 483: This "if" section got duplicated.

schlafly (commented):
This looks good.  Can you add a section to the documentation with maybe your image and a little bit of information?  i.e.:
- density of sources
- method for simulating Roman fluxes (linear interpolation among ultra vista wavelengths)
- depth of simulation (incomplete beyond mag ~ 25 IIRC?---not super deep)
- basic simulation parameters (random angles, flat distribution of sersic indices, sizes from COSMOS HST imaging)
- ... something else I'm forgetting?
- real Gaia stars with approximate fluxes (no real colors)

Re the density, can you remind me if you think it is right?  Right now you compute the area as (ramax - ramin) * (decmax - decmin), and then get the density as the number divided by the area.  I worry that the ultra vista area is only a portion of the whole field, and so (ramax - ramin) * (decmax - decmin) isn't a good approximation of the area that you end up keeping after the cuts.  Did you check that?

Can you also say what ends up in the "streamlined" catalog?  The size is reasonable so it's okay from that perspective, but we should probably ask permission before providing it as an alternative source of the COSMOS catalog unless it's sufficiently stripped down as to have limited scientific value beyond our use case.  We should also cite the COSMOS papers extensively in your new docs.

PaulHuwe:
The streamlined catalog has had the following cuts from the full COSMOS catalog:
- Trimmed to only the columns we need: 'FLUX_RADIUS', 'DELTA_J2000', 'UVISTA_H_FLUX_AUTO', 'HSC_r_FLUX_AUTO', 'HSC_z_FLUX_AUTO', 'lp_type','UVISTA_Y_FLUX_AUTO', 'ID', 'ALPHA_J2000', 'UVISTA_J_FLUX_AUTO', 'UVISTA_Ks_FLUX_AUTO', 'ACS_A_WORLD', 'FLAG_UDEEP', 'ACS_B_WORLD'
- Only keep 'lp_type" galaxies and Xray sources
- Trim to only those that contain the ultra-deep regions of UltraVISTA data
- Keep those with positive, nonzero flux radius, shape measurements (ACS_A_WORLD, ACS_B_WORLD )

These cuts trim the total number of objects from 1720700 to 345994. The filters in particular are very Roman-specific.

romanisim/catalog.py
PaulHuwe: Removed.

docs/romanisim/catalog.rst
```diff
@@ -31,6 +31,36 @@ Following these required fields is a series of columns giving the fluxes of the
...
+The simulator API includes a few simple tools to generate parametric distributions of stars and galaxies.  
+
+The ``make_stars`` and ``make_galaxies`` routines make random catalogs of stars and galaxies.  The number of stars and galaxies can be adjusted.  Likewise, the power law index by which the sources' magnitudes are sampled can be adjusted, as can their limiting magnitudes.  Galaxy Sersic parameters, half-light radii, and position angles are chosen at random, with a rough attempt to make brighter galaxies appropriately larger (i.e., conserving surface brightness).  Stars can be chosen to be distributed with a King profile.  This functionality is however very rudimentary and limited, and is better suited for toy problems than real scientific work.  
+
+We expect scientific uses to be driven by custom-created catalogs rather than the simple routines above, and provide the ``make_cosmos_galaxies`` and ``make_gaia_stars`` routines outlined below to provide that funcionality.
```
schlafly, line 38:
```suggestion
We expect scientific uses to be driven by custom-created catalogs rather than the simple routines above.  The routines ``make_cosmos_galaxies`` and ``make_gaia_stars`` outlined below provide some more scientific realism and may be adequate for some use cases.
```

docs/romanisim/catalog.rst
```diff
@@ -31,6 +31,36 @@ Following these required fields is a series of columns giving the fluxes of the
...
+ 
+The simulator performs the following operations on the catalog sources:
+
+* Only galaxies that have "ultra-deep" Ultra Vista wavelength measurements are selected from the COSMOS catalog. This trims the catalog to 0.62 square degrees, and to a peak magnitude depth of ~25. 
+* The source density of the COSMOS catalog as well as the input viewing area are used to determine the number of sources simulated.
+* Roman bands are approximated by linear interpolations of nearest HST and Ultra Vista bands.
+* Galaxies are placed in random positions about the sky.
+* While galaxy shapes are preserved, they are given randomly chosen concentrations (1 <= n <= 4).
+* Sources are randomly oriented.
+* Source fluxes are lightly perturbed.
```
schlafly, line 60:
```suggestion
* Source fluxes and colors are lightly perturbed.
```

docs/romanisim/catalog.rst
```diff
@@ -31,6 +31,36 @@ Following these required fields is a series of columns giving the fluxes of the
...
+
+The simulator can utilize the COSMOS catalog to generate sources based off of real galaxies via the ``make_cosmos_galaxies`` method. The COSMOS survey is a 2 square degree collection of over 2 million galaxies with multi-wavelength photometry (Weaver et al. 2022)[https://arxiv.org/pdf/2110.13923]. A streamlined catalog for the simulator is provided with the code, but for those so interested, the latest full catalog can be found at https://cosmos.astro.caltech.edu/ and used with this method. 
+ 
+The simulator performs the following operations on the catalog sources:
+
+* Only galaxies that have "ultra-deep" Ultra Vista wavelength measurements are selected from the COSMOS catalog. This trims the catalog to 0.62 square degrees, and to a peak magnitude depth of ~25. 
+* The source density of the COSMOS catalog as well as the input viewing area are used to determine the number of sources simulated.
+* Roman bands are approximated by linear interpolations of nearest HST and Ultra Vista bands.
+* Galaxies are placed in random positions about the sky.
+* While galaxy shapes are preserved, they are given randomly chosen concentrations (1 <= n <= 4).
```
schlafly, line 58:
```suggestion
* While galaxy sizes and ellipticities are preserved, they are given randomly chosen concentrations (1 <= n <= 4), and detailed shape information is ignored.
```

docs/romanisim/catalog.rst
```diff
@@ -31,6 +31,36 @@ Following these required fields is a series of columns giving the fluxes of the
...
+  :width: 600
+
+  Level 3 Mosaic of cosmos and gaia sources with radius = 0.1 degree (search area visible)
+
+
+The simulator can utilize the COSMOS catalog to generate sources based off of real galaxies via the ``make_cosmos_galaxies`` method. The COSMOS survey is a 2 square degree collection of over 2 million galaxies with multi-wavelength photometry (Weaver et al. 2022)[https://arxiv.org/pdf/2110.13923]. A streamlined catalog for the simulator is provided with the code, but for those so interested, the latest full catalog can be found at https://cosmos.astro.caltech.edu/ and used with this method. 
+ 
+The simulator performs the following operations on the catalog sources:
+
+* Only galaxies that have "ultra-deep" Ultra Vista wavelength measurements are selected from the COSMOS catalog. This trims the catalog to 0.62 square degrees, and to a peak magnitude depth of ~25.
```
schlafly, line 54: We should think about how to go fainter than 25th, which is not very deep for Roman.  e.g., this might look like:
- fit a power law to the number distribution as a function of magnitude for H < 24 or so.
- draw a number from that power law.
- If the number corresponds to H < 24, use the current approach.
- If the number corresponds to H > 24, use the current approach but then change all the magnitudes of the selected galaxy by a constant equal to H_drawn - H_selected galaxy.
That will screw up the colors as a function of redshift but will get large numbers of faint galaxies, which we'd like.  Maybe think about how much work such an implementation would be, and consider it as an option if it works?

romanisim/catalog.py
```diff
@@ -241,6 +442,73 @@ def make_galaxies(coord,
...
+    out = table.Table()
+    out['ra'] = star_cat['ra']
+    out['dec'] = star_cat['dec']
+    out['type'] = types
+    out['n'] = -1 * np.zeros(len(star_cat), dtype='f4')
+    out['half_light_radius'] = np.zeros(len(star_cat), dtype='f4')
+    out['pa'] = np.zeros(len(star_cat), dtype='f4')
+    out['ba'] = np.ones(len(star_cat), dtype='f4')
+    for bandpass in bandpasses:
+        out[bandpass] = star_cat[bandpass]
```
schlafly, line 511: What is this function doing on top of returning star_cat?  I'm missing the difference between out and star_cat.

The docs reference "false color" stars but gaia2romanisimcat just assumes color = 0 in all possible bands.  I briefly thought when you said that you might have meant that you were using Gaia BP/RP to predict the Roman bands, but that's not what's happening.  I've made a suggestion for the docs there.

docs/romanisim/catalog.rst
```diff
@@ -31,6 +31,36 @@ Following these required fields is a series of columns giving the fluxes of the
...
+* The source density of the COSMOS catalog as well as the input viewing area are used to determine the number of sources simulated.
+* Roman bands are approximated by linear interpolations of nearest HST and Ultra Vista bands.
+* Galaxies are placed in random positions about the sky.
+* While galaxy shapes are preserved, they are given randomly chosen concentrations (1 <= n <= 4).
+* Sources are randomly oriented.
+* Source fluxes are lightly perturbed.
+
+The simulator allows for passage of a user-specified COSMOS file (one which must have the same tabular format as COSMOS2020_CLASSIC_R1_v2.2_p3.fits) via the ``filename`` keyword. The catalog area can be set with ``cat_area``.
+
+The simulator can also generate the appropriate false-color stars for the image from the GAIA catalog via the ``make_gaia_stars`` method, in place of the random stars in ``make_stars``.
```
schlafly, line 64: false-color to me implies some attempt to get the star colors right, but this follows gaia.gaia2romanisimcat that just does AB color = 0 in all bands.

```suggestion
The simulator can also generate place Gaia stars at the appropriate locations using the ``make_gaia_stars`` method, in place of the random stars in ``make_stars``.  Note that these stars have fluxes exactly equal to the Gaia G band fluxes in all bands; i.e., no attempt is made to predict accurate colors of stars.
```

schlafly (commented):
This looks good.  Can I ask you to make one 3-color image using your new catalog?  I'd simulate an image in each of three bands, read each of the images into memory, and use this function
https://github.com/spacetelescope/romanisim/blob/main/romanisim/util.py#L74
to try to get a good scaling.  Mostly it would be good to see that the colors look kind of realistic.  It would be good to replace the current black/white image with this one.

docs/romanisim/catalog.rst
PaulHuwe: I was thinking about trying to base it off of the mag density of the HST observations and extrapolating from there. I think it is worth further discussion.

docs/romanisim/catalog.rst
PaulHuwe: Changed

docs/romanisim/catalog.rst
PaulHuwe: Changed

docs/romanisim/catalog.rst
PaulHuwe: Changed

docs/romanisim/catalog.rst
PaulHuwe: Added this to the JIRA board

docs/romanisim/catalog.rst
PaulHuwe: Changed

schlafly:
Image looks good, thanks.  Can you help me with
https://github.com/spacetelescope/romanisim/pull/199#discussion_r1977969097
?

romanisim/catalog.py
PaulHuwe: I did not realize that star_cat was already in the correct format. I will streamline this.

schlafly: Thanks.  The first part of this function looks reasonable but it duplicates
https://github.com/spacetelescope/romanisim/blob/main/scripts/romanisim-gaia-catalog#L28-L33
It's fine to have it here instead but then let's refactor romanisim-gaia-catalog to use it.  Looking more at this, it looks like both of us ended up writing romanisim-gaia-catalog / romanisim-make-catalog methods.  Can you make sure that both romanisim-make-catalog and romanisim-gaia-catalog do the same thing and then merge / get rid of one as part of this?  It looks like we made mildly different decisions about what arguments are required vs. not but the two modules are pretty identical.

PaulHuwe: Addressed (both the cat question and the docs update).

PaulHuwe: I streamlined my GAIA bit to nto have the catalog duplication. The function needs to have the code from the script, as the script is not important - however the script could be changed to use the catalog function. 

Do you want me to update the script in this ticket or a new one?

schlafly: Your call; I just want to eliminate the code duplication.

schlafly (approved):
Merge when ready, thanks!

PaulHuwe:
I have removed the duplicate code in romanisim-gaia-catalog and confirmed the output remained effectively unchanged.

PR #201, opened 2025-02-11
RCAL-1015: Failing B17 Tests Hotfix

PaulHuwe:
A hot fix for the tests presently failing.


schlafly:
Any word on when we expect this to be fixed in gwcs or if there's a workaround?  Should I go ahead and pin in the interim?  The only reason to avoid the pin would be to try to make romanisim more compatible with others' environments.

PR #203, opened 2025-02-14
fix webbpsf data download cache

braingram:
Similar to https://github.com/spacetelescope/romancal/pull/1619

This won't work until:
- it's merged
- the cache is regenerated (by running https://github.com/spacetelescope/romanisim/actions/workflows/data.yml)

This is a stop-gap fix until this package can be migrated to stpsf.

### DIFF
--- a/.github/workflows/data.yml
+++ b/.github/workflows/data.yml
@@ -33,7 +33,7 @@ on:
 
 jobs:
   download_webbpsf_data:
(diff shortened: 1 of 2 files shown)


braingram:
Thanks! I don't have permissions to trigger the workflow at:
https://github.com/spacetelescope/romanisim/actions/workflows/data.yml
There should be a button to "Run workflow" (all defaults are fine)
Screenshot 2025-02-14 at 11 42 55 AM

PR #205, opened 2025-02-17
Migrate to stpsf.

schlafly:
This PR replaces WebbPSF with STPSF, preserving only the --webbpsf argument (with a deprecation warning) to the main script.


schlafly:
@braingram , this worked, contrary to my expectations that I would need to do something new to get the new workflows running or something.  Any reason I shouldn't merge this?  Do I want to call the stpsf workflow rather than the vendorized one here?

scripts/romanisim-make-image
```diff
@@ -98,7 +98,9 @@ if __name__ == '__main__':
     parser.add_argument('--usecrds', action='store_true',
                         help='Use CRDS for distortion map')
     parser.add_argument('--webbpsf', action='store_true',
-                        help='Use webbpsf for PSF')
+                        help='Use stpsf for PSF')  # deprecated!
```
braingram, line 101:
```suggestion
                        help='Use stpsf for PSF (deprecated, use stpsf)')
```

scripts/romanisim-make-stack
```diff
@@ -94,11 +94,16 @@ def main():
                         help='Use CRDS for distortion map')
 
     # Use CRDS stored PSFs
+    # Deprecated!
     parser.add_argument('-w', '--webbpsf',
                         action='store_true',
-                        help='Use webbpsf for PSFs.')
+                        help='Use stpsf for PSFs.')
```
braingram, line 100:
```suggestion
                        help='Use stpsf for PSFs. (deprecated, use stpsf)')
```

braingram (commented):
I also don't trust the green :)

Turns out there was a silent "ERROR" (which looks to be a warning):
https://github.com/spacetelescope/romanisim/actions/runs/13375395452/job/37353442874?pr=205#step:10:211

I think part of the migration will involve updating:
https://github.com/spacetelescope/romanisim/blob/bc586c79d240979d44ac9a08f366f2626a26b58a/.github/workflows/download_data.yml#L67-L68
to point to whatever the new urls are for stpsf. Otherwise looking through this PR it looks like you caught most (if not all the rest) of the required changes.

docs/romanisim/install.rst
```diff
@@ -44,12 +44,12 @@ In summary, the basic install process looks like this::
...
-    # get webbpsf data and untar it
-    mkdir -p $HOME/data/webbpsf-data
-    cd $HOME/data/webbpsf-data
-    wget https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz -O webbpsf-data.tar.gz
-    tar -xzf webbpsf-data.tar.gz
-    export WEBBPSF_PATH=$PWD/webbpsf-data
+    # get stpsf data and untar it
+    mkdir -p $HOME/data/stpsf-data
+    cd $HOME/data/stpsf-data
+    wget https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz -O stpsf-data.tar.gz
```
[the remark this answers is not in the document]
schlafly: I've followed your advice and pointed to the documentation instead.

schlafly:
Thanks.  I've updated the data directory links but still have the same missing STPSF data warnings.  I presume that's because I need to do something like merge this PR and then run the new action in order to get the new cache for this to look at?  So the right procedure is to merge, run the workflow, and cross my fingers?

romanisim has the problem that it has no regression tests, so we don't need any updates there.

The refactoring you're referring to is roughly moving the workflow back to use the stpsf workflow instead of a custom workflow in each product?

Maybe I should turn warnings into errors but I'll skip that for now.

braingram:
> Thanks. I've updated the data directory links but still have the same missing STPSF data warnings. I presume that's because I need to do something like merge this PR and then run the new action in order to get the new cache for this to look at? So the right procedure is to merge, run the workflow, and cross my fingers?
> 
> romanisim has the problem that it has no regression tests, so we don't need any updates there.
> 
> The refactoring you're referring to is roughly moving the workflow back to use the stpsf workflow instead of a custom workflow in each product?
> 
> Maybe I should turn warnings into errors but I'll skip that for now.

Honestly the whole thing is confusing. Now the existing box links have been updated to point to stpsf (which is causing the romancal regtests to fail).

I say we yolo these changes and see if it helps.

braingram (approved):

schlafly:
It looks like the new stpsf data files still unzip to a webbpsf directory, which meant that this attempt didn't work.
https://github.com/spacetelescope/romanisim/actions/runs/13379317455/job/37364941998#step:4:1

schlafly:
No, it looks like another file does unzip to an stpsf directory... investigating.

braingram:
I think the commit you pushed to fix the url should do it. It looks like the latest data run generated a cache with key:
`data-stpsf-data-mini-2.0.0-galsim-data-1e8e65479e194d5c8a2e16857004de8b46ed1829f9fc11b7697dd7c7126670a1`
(I don't have permissions to see the caches).
However the latest CI run didn't pick up that key (perhaps it started before the cache was made):
https://github.com/spacetelescope/romanisim/actions/runs/13379441148/job/37365284833
When you have a chance if you would re-trigger the CI workflow we can see if it all works again.

schlafly:
I haven't figured this out.  In the run you linked, it looks like somehow the key name got mangled and lost the "-2.0.0-" component.
https://github.com/spacetelescope/romanisim/actions/runs/13379441148/job/37365284833#step:3:11
But most of the CI runs I've done have seem to me to have exactly the right cache key but still fail.
https://github.com/spacetelescope/romanisim/actions/runs/13381621082/job/37371034099#step:3:11
So I don't understand how it got mangled in that one example or why it's still failing when the key looks good.

schlafly:
PS: I went ahead and added you to romanisim-maintainers so you could poke around better, but let me know and I'll remove you again---I don't think you need any more email!

PR #207, opened 2025-02-24
test with latest supported Python version

zacharyburnett:
[SCSB-195](https://jira.stsci.edu/browse/SCSB-195)


schlafly:
Thanks!

PR #208, opened 2025-03-07
Improve scalergb to make nicer images.

schlafly:
This swaps out the existing scalergb solution for one derived from the Legacy Survey viewer, which in turn was derived from the SDSS viewer, and updates the tests accordingly.

### DIFF
--- a/romanisim/util.py
+++ b/romanisim/util.py
@@ -71,15 +71,11 @@ def celestialcoord(sky):
                                      sky.dec.to(u.rad).value * galsim.radians)
 
 
-def scalergb(rgb, scales=None, lumrange=None):
-    """Scales three flux images into a range of luminosity for displaying.
+def scalergb(rgb, scales=None, gray=0):
+    """Scales three flux images into an RGB image for displaying.
 
-    Images are scaled into [0, 1].
-
-    This routine is intended to help with cases where you want to display
-    some images and want the color scale to cover only a certain range,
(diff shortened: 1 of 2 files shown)


romanisim/util.py
```diff
@@ -71,44 +71,38 @@ def celestialcoord(sky):
...
-        lumrange = [0, np.max(norm)]
-    newnorm = np.clip((norm - lumrange[0]) / (lumrange[1] - lumrange[0]),
-                      0, 1)
-    out = rgb.copy()
-    for i in range(3):
-        out[:, :, i] = out[:, :, i] * newnorm / (norm + (norm == 0))
-    return out
+            rgb[:, :, i] = rgb[:, :, i] / scales[i] + gray
+    intensity = np.sum(np.clip(rgb, 0, np.inf), axis=2)  # how much light is there?
+    Q = 20  # taken from legacyviewer
```
PaulHuwe, line 100: Maybe make Q a function parameter so it can be changed, if desired?

PaulHuwe (approved):
One possible change suggested, otherwise LGTM

romanisim/util.py
schlafly: I thought about it, but I have poor intuition for Q, and if it hasn't been promoted to a real argument in the Legacy Survey code base it seems unlikely to me that we'll end up changing it here.

PR #210, opened 2025-03-28
Updates for putting filters in filenames in romanisim.

schlafly:
The Roman L1 & L2 file naming scheme was changed to include the filter name in lowercase in the file name.  This PR implements that change in the romanisim regtest file creation routines.  It also tries to make a more general routine for "automatically" getting the bandpass in the filename, supporting filename arguments like
romanisim-make-image out-{}-{bandpass}.asdf
to produce out-wfi01-f087.asdf.  The naming for the SCA argument should have been {sca} in retrospect, but we retain the old API that was used to support
romanisim-make-image out-{}.asdf --sca -1
to trigger generating all 18 SCAs with names like out-wfi01.asdf.

This PR also adds back MA tables 109 and 110.  These are not officially supported as part of the PRD, so they were removed.  But they have spectroscopic & imaging modes and 110 has >= 16 resultants, meaning that they enable DMS testing in a way that the MA tables in the PRD do not (no >= 16 resultant MA tables for imaging, different MA table numbers for imaging & spectroscopy).  We probably need a better way of dealing with that in general but this approach works for now.

An unrelated change to romanisim-make-catalog slipped in allowing romanisim-make-catalog to write out files in formats other than ecsv.

### DIFF
--- a/romanisim/parameters.py
+++ b/romanisim/parameters.py
@@ -153,7 +153,23 @@ read_pattern = {3: [[2], [5], [10], [16], [17]],
                      [95 + x for x in range(32)], 
                      [127 + x for x in range(32)], 
                      [159 + x for x in range(32)], 
-                     [191 + x for x in range(32)], [223]]
+                     [191 + x for x in range(32)], [223]],
+                18: [[x] for x in range(1, 301)],
+
+                # note: these MA tables are not part of the PRD and are intended
+                # only to support DMS testing.  In particular, 110 is a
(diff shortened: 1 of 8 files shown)


PaulHuwe (approved):
LGTM - I presume you checked that the produced output properly validates?

schlafly:
Yeah, I built a set of regtest files with this and updated romancal to use them; I'm not worried on that end.

PR #214, opened 2025-04-22
Bug Fix: Hardcode path for default cosmos catalog

stscieisenhamer:
While testing, noticed that `test_catalog` was failing finding the default COSMOS catalog when `pytest` was not run from the package folder. Traced to the hardcoded definition in `catalog.py` where the path to the catalog was left just as a relative path.

### DIFF
--- a/romanisim/catalog.py
+++ b/romanisim/catalog.py
@@ -3,6 +3,8 @@
 This module provides basic routines to allow romanisim to render scenes
 based on catalogs of sources in those scenes.
 """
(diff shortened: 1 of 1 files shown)


schlafly:
William had a related PR from last week that I just merged.  Does that PR fix your issue?  https://github.com/spacetelescope/romanisim/pull/213

stscieisenhamer:
yep, that will do it. this one can be closed out.

PR #223, opened 2025-06-18
Fix amp33 and wcs to allow pipeline processing

schlafly:
The migration away from maker_utils ended up having (at least) two unintended consequences:
- amp33 was getting populated as a [n_resultants, 4096, 4096] array instead of the usual [n_resultants, 4096, 128] array.  This inflates the size of the L1 files and leads to breakage in the reference pixel correction step.  We didn't notice this since romanisim itself doesn't actually do anything with amp33.
- mk_level2_image was making a data array with a 4k x 4k shape, but create_fake_data was leaving this 0x0 by default.  The WCS creation was using the size in setting the bounding box, which was leading to a 0x0 bounding box in some cases.

This PR fixes those two issues.

### DIFF
--- a/romanisim/l1.py
+++ b/romanisim/l1.py
@@ -412,6 +412,8 @@ def make_asdf(resultants, dq=None, filepath=None, metadata=None, persistence=Non
     nborder = parameters.nborder
     npix = galsim.roman.n_pix + 2 * nborder
     out = stnode.WfiScienceRaw.create_fake_data(shape=(len(resultants), npix, npix))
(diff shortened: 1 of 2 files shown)


braingram (approved):
Thanks! Looks good to me. Sorry for the complication with this API change.

braingram:
I queued up a run with this PR at: https://github.com/spacetelescope/RegressionTests/actions/runs/15740191800

PaulHuwe (approved):
LGTM

PR #231, opened 2025-07-21
L3 updates

schlafly:
We updated the L3 metadata for romancal for the intermediate release.  This brings romanisim in line with those updates.


braingram (approved):
Thanks! LGTM.

PR #236, opened 2025-08-06
Remove gwcs upper pin

schlafly:
We had a PR in place to address an issue with gwcs 0.24.0.  We now require gwcs >=0.25.0 and remove this pin.


PaulHuwe (approved):
LGTM

PR #241, opened 2025-08-18
Use latest roman-datamodels release.

mairanteodoro:
Adding `romanisim` as a test dependency for `roman-photoz` was installing a version of `roman-datamodels` that was different from what was needed to be compatible with `rad` schema. This PR points `roman-datamodels` to the latest release instead of the dev version.

Not 100% sure whether we can always use the latest `roman-datamodels` release, but this fixed all the issues in `roman-photoz` and we were able to run the scripts and produce viable results.


pyproject.toml
```diff
@@ -3,54 +3,50 @@ name = "romanisim"
...
-    "stpsf >=2.1.0",
-    "Cython >=0.29.21",
+  "asdf >=4.1.0",
+  "astropy >=6.0",
+  "astropy_healpix >= 1.1.2",
+  "crds >=11.16.21",
+  "defusedxml >=0.5.0",
+  "galsim >=2.5.1",
+  # "roman_datamodels>=0.26.0,<0.27.0",
+  "roman_datamodels",
```
PaulHuwe, line 24:
```suggestion
  "roman_datamodels>=0.27.0",
```

PaulHuwe (changes requested):

PaulHuwe:
I think this can be updated, resolved, and merged?

schlafly (approved):

PR #246, opened 2025-08-28
remove python upper pin

braingram:
Follow-up to https://github.com/spacetelescope/romanisim/pull/235

This PR will stay draft until all our dependencies are 3.14 compatible (and perhaps when we also have 3.14 testing).

Rebase, push, re-run as needed to test with 3.14.


braingram:
The 3.14 build is still failing due to galsim:
https://github.com/GalSim-developers/GalSim/issues/1339

PaulHuwe (approved):
LGTM - It is good to finally release this restriction!

PR #247, opened 2025-08-28
Update CRDS regtest to trigger CRDS ma table reference storage.

schlafly:
#224 caused one of the regression tests to fail because that test was incompletely setting up the metadata needed to run files with CRDS now.  This PR addresses that issue.


PaulHuwe:
This passed regression testing:
https://github.com/spacetelescope/RegressionTests/actions/runs/17365945784

PR #256, opened 2025-09-12
Ability to add in extra_counts

npadmana:
The following hooks just allow passing in an additional array to be added to the simulation. This now just needs to get added  into the main script.

### DIFF
--- a/scripts/romanisim-make-image
+++ b/scripts/romanisim-make-image
@@ -9,6 +9,8 @@ import galsim
 from romanisim import log, wcs, persistence, parameters
 from romanisim import ris_make_utils as ris
 from copy import copy
+import numpy as np
+from astropy.io import fits
 
 
 def go(args):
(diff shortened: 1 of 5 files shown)


scripts/romanisim-make-image
```diff
@@ -58,8 +60,27 @@ def go(args):
...
+            extra_counts_file = args.extra_counts[0]
+            extra_counts_hdu = int(args.extra_counts[1])
+        else:
+            raise ValueError("If --extra-counts is used, must provide one or two arguments only.")
+        # Undocumented, but useful for testing 
+        if extra_counts_file=="_testing":
+            extra_counts = np.zeros((4088, 4088), dtype=np.float32)+float(extra_counts_hdu)
+        else:
+            with fits.open(extra_counts_file) as hdul:
+                extra_counts = hdul[extra_counts_hdu].data
```
schlafly, line 79:
```suggestion
        with fits.open(extra_counts_file) as hdul:
            extra_counts = hdul[extra_counts_hdu].data  
```

My inclination is to delete this.  If it got Poisson sampled it could be a hacky mechanism for changing the sky background & noise, but as is I have trouble seeing when people might want to use this.

npadmana: It's only here for me to test that the code runs, without having to create an extra file. Which is why I tried to bury it with an obscure call. 

But I'm happy to delete this, if you'd prefer.

schlafly (approved):
Perfect, thanks.

PR #258, opened 2025-09-18
Updates to support GRISM/PRISM bandpasses

npadmana:
Changes include -

- Adding to the roman->galsim translations
- STPSF has two grism options, we default to GRISM1
- Renaming column headers in effective area files
- Also make the changes in --pretend_spectral
- Fix up the exposure time to use the `frame_time` information in the MA Table if available...

Note that you need to use the `--stpsf` option here.

This allows for calls like 

```
romanisim-make-image out.asdf --usecrds --radec 270 66 --extra-counts extra_counts.fits --ma_table_number 1036 --bandpass GRISM --nobj 0 --stpsf
```

### DIFF
--- a/romanisim/catalog.py
+++ b/romanisim/catalog.py
@@ -244,6 +244,14 @@ def make_cosmos_galaxies(coord,
                 cos_filt.append('UVISTA_Ks_FLUX_AUTO')
             if opt_elem in ("F158", "F184", "F146"):
                 cos_filt.append('UVISTA_H_FLUX_AUTO')
+            # Mirror the GRISM/PRISM mapping used below when assigning
+            # FLUX_GRISM/FLUX_PRISM, so the cosmos catalog read pulls a
+            # column with non-zero entries and source filtering still leaves
+            # rows behind.
+            if opt_elem == "GRISM":
+                cos_filt.append('UVISTA_H_FLUX_AUTO')
+            if opt_elem == "PRISM":
+                cos_filt.append('UVISTA_J_FLUX_AUTO')
     cos_filt = list(set(cos_filt))
 
     # Open COSMOS file and pare to required tabs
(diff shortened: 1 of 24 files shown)


schlafly:
Re the failing tests, we make some ~COSMOS-derived catalogs by taking fluxes of real objects from COSMOS and doing some rough color transforms to get approximate Roman colors in the various bands.  The new grism / prism filters don't have that and don't work.  I think if you make up a "color transformation" there, e.g., setting PRISM = UVISTA_J_FLUX_AUTO, GRISM = UVISTA_H_FLUX_AUTO, then that routine will be able to invent prism and grism fluxes.  That's probably easier than special casing the tests and isn't crazy.

npadmana:
Hi @schlafly — here is an updated version of this PR. Sorry for how long this has taken. Hopefully, tests pass here.

One comment — this introduces the GRISM and PRISM bandpasses. For now, I think this is really only useful for people who want to use the extra_counts, so these are being used to estimate the sky background. I think this will be a somewhat optimistic, since it is using eg the GRISM 1st order bandpass, but really, all the light in the grism wavelength range should end up on the detector. It may also not be uniform. But this is a first pass.

npadmana:
Quick note — the extra_counts now accepts an optional HDU number, or a name for a cleaner interface.

schlafly:
This looks good.  One important question: you have updated 18 files in roman-technical-information/data.  These files are direct copies from https://github.com/RomanSpaceTelescope/roman-technical-information/, and are 'fall backs' if roman-technical-information isn't available otherwise.  Accordingly, it seems bad to edit them.  Is there a good mechanism to avoid that?

Otherwise this PR looks good and I would merge.

npadmana:
@schlafly -- done. Put in a translation specific to reading the ECSV files, so should not interfere with other code. Reverted the tech files. Tests passed locally....

schlafly (approved):

schlafly:
Thank you!

PR #261, opened 2025-09-30
RCAL-1107 Create simulations based psfs from the CRDS EPSF reftype

stscieisenhamer:
Resolves [RCAL-1107](https://jira.stsci.edu/browse/RCAL-1107)

Previously, PSF's could be determined using the `galsim` package, by default, or using the `stpsf` package. A third PSF determination method has been added, called `epsf`. The EPSF reference reftype from CRDS is used to retrieve empirical PSF's to use in the simulation calculations.

The command-line changes include:

- Addition of the option `--psftype` to choose between all three methods
- Deprecation of the `--stpsf` and `--webpsf` options. The options are still avialable, but come with deprecation warnings.

### DIFF
--- a/romanisim/psf.py
+++ b/romanisim/psf.py
@@ -426,8 +426,8 @@ def psf_from_grid(psfgrid, x_0=None, y_0=None, size=185):
     if size % 2 == 0:
         raise ValueError(f'Argument `size` is required to be odd. Given: {size}')
 
-    x_0 = 0. if x_0 is None else x_0
-    y_0 = 0. if y_0 is None else y_0
+    x_0 = 2048 if x_0 is None else x_0
+    y_0 = 2048 if y_0 is None else y_0
     cc = (np.arange(size) - (size // 2)) / psfgrid.meta['epsf_oversample']
     x, y = np.meshgrid(cc + x_0, cc + y_0)
     psf = psfgrid.evaluate(x, y, 1, x_0, y_0)


romanisim/psf.py
```diff
@@ -13,16 +13,139 @@
...
+        psf = psf_images[i, :, :]
+        im = convolve(psf, pixel_response_kernel) * oversample**2
+        psf_images[i, :, :] = im
+
+    meta["grid_xypos"] = position_list
+    meta["oversampling"] = oversample
+    nd = NDData(psf_images, meta=meta)
+    model = GriddedPSFModel(nd)
+
+    return model
```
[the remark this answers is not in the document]
stscieisenhamer: In reference to `GriddedPSFModel`? Not immediately but can determine. If it is setting up the interpolation fit at initiation, it might have a little bit of heft, but I would think not significant.

If the concern is the "recreating the basic model every time a psf is requested, through the `make_psf` func". A very easy/simple solution is used where the inner funcs `get_gridded_psf_model` and `get_epsf_from_crds` are wrapped in the `@cache` decorator. For any single configuration of sca/filter, these routines, though called multiple times, will only actually be executed once.

romanisim/tests/test_psf.py
```diff
@@ -20,32 +21,36 @@ def local(self, *args, **kwargs):
...
-        # assert that image catches no more than 100% and no less than 90%
-        # of flux?
-        assert np.min(im) > np.max(im) * (-1e-3)
-        # ideally nothing negative
+
+    if not kwargs.get('chromatic', False):
+        method = 'auto'
+        if kwargs.get('psftype', None) in ('epsf', 'stpsf'):
+            method = 'real_space'
+        im = p.drawImage(method=method).array
```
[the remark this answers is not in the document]
stscieisenhamer: Yes, definitely needs discussion. Was on my list.

scripts/romanisim-make-l3
```diff
@@ -31,7 +31,7 @@ if __name__ == '__main__':
     parser.add_argument('--date', type=str, default=None,
                         help=('UTC Date and Time of observation to simulate in ISOT format.'))
     parser.add_argument('--radec', type=float, nargs=2,
-                        help='ra and dec (deg)', default=None)
+                        help='ra and dec (deg)', default=(270, 66))
```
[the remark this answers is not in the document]
stscieisenhamer: Will look. I have rebased and this remains so the fix is somewhere else maybe?

scripts/romanisim-make-stack
```diff
@@ -93,15 +93,18 @@ def main():
                         action='store_true',
                         help='Use CRDS for distortion map')
 
-    # Use CRDS stored PSFs
-    # Deprecated!
+    parser.add_argument('--psftype', nargs=1, type=str, default=None,
```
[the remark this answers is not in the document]
stscieisenhamer: That is a mistake and missed this one occurrence. `nargs` specification should not be there.

romanisim/psf.py
```diff
@@ -13,16 +13,139 @@
...
+        Date of simulation. If None, current date is used.
+
+    Returns
+    -------
+    model : roman_datamodels.EpsfRefModel
+    """
+    from crds import getreferences
+
+    if date is None:
+        date = Time.now()
```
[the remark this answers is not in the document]
schlafly: The way we've done this so far is to put something in romanisim.parameters.py.  Right now there are three default dates in there that are all 2026-01-01.  Those should probably be pulled out into parameters.default_date and then used elsewhere.

My aversion to Time.now() is that then you get different behavior each time you run it if you don't notice it.

I'm okay with making it a required argument if you'd rather.

scripts/romanisim-make-l3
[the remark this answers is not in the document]
schlafly: Sorry, yes, by separately I meant that I made whatever function it was that wasn't handling None right handle it better, in my limited recall!

romanisim/psf.py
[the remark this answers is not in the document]
stscieisenhamer: Ah, a configuration file parameter. That is good.

romanisim/psf.py
```diff
@@ -171,49 +390,123 @@ def make_psf(sca, filter_name, wcs=None, stpsf=True, pix=None,
...
+        n_y = ceil(y_end - y_start)
+    rx = np.array([np.linspace(x_start, x_end, n_x)])
+    x = rx.repeat(n_y, axis=0)
+    ry = np.array([np.linspace(y_start, y_end, n_y)]).T
+    y = ry.repeat(n_x, axis=1)
 
-    def __init__(self, corners, psf):
-        self.corners = corners
-        self.psf = psf
+    return x, y
```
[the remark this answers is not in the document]
stscieisenhamer: Not modified by me, but can replace.

romanisim/psf.py
```diff
@@ -171,49 +390,123 @@ def make_psf(sca, filter_name, wcs=None, stpsf=True, pix=None,
...
+    Returns
+    -------
+    psf : nd.array
+        The psf profile.
+    """
+    x_0 = 0. if x_0 is None else x_0
+    y_0 = 0. if y_0 is None else y_0
+    bb = psfgrid.get_bounding_box()
+    x, y = pix_coords(x_0 + bb.intervals[0].lower, x_0 + bb.intervals[0].upper,
+                      y_0 + bb.intervals[1].lower, y_0 + bb.intervals[1].upper)
```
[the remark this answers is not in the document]
stscieisenhamer: Yes absolutely! I just did `pix_coords` because I couldn't remember/find the meshgrid solution. From previous discussion, its on the list to refactor.

stscieisenhamer:
Ignoring the regression creation scripts, I have noted that there is quite the mix of defaults for `psftype`, that were based on the original defaults derived from the old --stpsf switch. The current mix is None, galsim, and epsf. Some questions:

- Should None even be allowed? I would say that for all func definitions, an explicit default should be used and remove None.
- If psftype is to have a default, what should it be? It seems all is leaning towards epsf, but want to know for sure.

schlafly:
I think you could just point it at the public CRDS servers and it would work.  Alternatively we can leave CRDS out of the github testing for now.  But yes, thanks for getting that sorted.

Re defaults, I'm okay with getting rid of None.  I think the default has to be galsim in the absence of direction, since it works without a CRDS server and without stpsf.

docs/romanisim/install.rst
```diff
@@ -17,8 +17,10 @@ operate.  See the `docs
 <https://stpsf.readthedocs.io/en/latest/installation.html#installing-the-required-data-files>`_
 for instructions on obtaining the relevant data files and pointing the
 ``STPSF_PATH`` environment variable to them.  This issue can be
-avoided by not setting the ``--stpsf`` argument, in which case
-``romanisim`` uses the GalSim modeling of the Roman PSF.
+avoided by not setting the ``--psftype=stpsf`` argument, in which case
+``romanisim`` uses the GalSim modeling of the Roman PSF. Alternatively, setting
+``--psftype=epsf`` will cause the use of CRDS EPSF reference files; see below
+for general CRDS usage.
```
schlafly:
```suggestion
``STPSF_PATH`` environment variable to them.  The simplest option for getting
things to run is to use ``--psftype=galsim`` which relies on the GalSim PSF
modeling.
```

(trying to focus on the simple installation here while not getting bogged down in different kinds of PSFs)

docs/romanisim/running.rst
```diff
@@ -15,43 +15,49 @@ The ``romanisim-make-image`` command line interface (CLI) has a number of argume
...
       --roll ROLL           Position angle (North towards YIdl) measured at the V2Ref/V3Ref of the aperture used. (default: 0)
-      --sca SCA             SCA to simulate. Use -1 to generate images for all SCAs; include {} in filename for this mode to indicate where the SCA number should be filled, e.g.
-                            l1_wfi{}.asdf (default: 7)
-      --usecrds             Use CRDS for distortion map (default: False)
-      --stpsf               Use stpsf for PSF (default: False)
+      --sca SCA             SCA to simulate. Use -1 to generate images for all SCAs; include {} in filename for this mode to indicate where the detector number should be filled, e.g.
+                            l1_{}.asdf (default: 7)
+      --usecrds             Use CRDS references. (default: False)
+      --psftype {None,epsf,galsim,stpsf}
+                            Type of PSF generator to use. If None and --usecrds, then "epsf" will be used otherwise "galsim" is used. (default: None)
```
schlafly:
```suggestion
                            Type of PSF to use. If None and --usecrds, then "epsf" will be used otherwise "galsim" is used. (default: None)
```

romanisim/tests/test_l3.py
```diff
@@ -240,9 +240,9 @@ def set_up_image_rendering_things():
     im = galsim.ImageF(100, 100, scale=0.11, xmin=0, ymin=0)
     filter_name = 'F158'
     sca = 1
-    impsfgray = psf.make_psf(sca, filter_name, stpsf=True, chromatic=False,
+    impsfgray = psf.make_psf(sca, filter_name, psftype='stpsf', chromatic=False,
                              nlambda=1)  # nlambda = 1 speeds tests
-    impsfchromatic = psf.make_psf(sca, filter_name, stpsf=False,
+    impsfchromatic = psf.make_psf(sca, filter_name, psftype='stpsf',
```
schlafly: Surprised this works!  This needs to be galsim for the chromatic PSF.

romanisim/tests/test_psf.py
```diff
@@ -20,32 +21,40 @@ def local(self, *args, **kwargs):
...
-    psfs.append(psf.make_psf(5, 'F184', pix=(1000, 1000), stpsf=False))
-    psfs.append(psf.make_psf(6, 'F184', pix=(1000, 1000), stpsf=True,
-                             **pkw))
-    psfs.append(psf.make_psf(7, 'F129', wcs=FakeWCS(), **pkw))
-    chromatic = [False] * 7
-    chromatic[3] = True
+@pytest.mark.parametrize("args, kwargs, position", [
+    ((1, 'F087'), {'psftype': 'stpsf', 'nlambda': 1}, None),
+    ((2, 'F184'), {'psftype': 'stpsf', 'nlambda': 1}, None),
+    ((3, 'F087'), {'psftype': 'epsf', 'nlambda': 1}, None),
```
schlafly: I think nlambda only makes sense for stpsf?

romanisim/tests/test_psf.py
```diff
@@ -20,32 +21,40 @@ def local(self, *args, **kwargs):
...
-        assert totsum > 0.9
-        # assert that image catches no more than 100% and no less than 90%
-        # of flux?
-        assert np.min(im) > np.max(im) * (-1e-3)
-        # ideally nothing negative
+
+    if not kwargs.get('chromatic', False):
+        method = 'auto'
+        # if kwargs.get('psftype', None) in ('epsf', 'stpsf'):
+        #     method = 'real_space'
```
schlafly: Delete?

romanisim/tests/test_psf.py
```diff
@@ -20,32 +21,40 @@ def local(self, *args, **kwargs):
...
+    ((4, 'F184'), {'psftype': 'galsim'}, None),
+    ((5, 'H158'), {'psftype': 'galsim'}, None),
+    ((6, 'F184'), {'pix': (1000, 1000), 'psftype': None}, None),
+    ((7, 'F184'), {'pix': (1000, 1000), 'psftype': 'galsim'}, None),
+    ((8, 'F184'), {'pix': (1000, 1000), 'psftype': 'stpsf', 'nlambda': 1}, None),
+    ((9, 'F184'), {'pix': (1000, 1000), 'psftype': 'epsf', 'nlambda': 1}, None),
+    ((10, 'F129'), {'psftype': 'stpsf', 'wcs': FakeWCS(), 'nlambda': 1}, None),
+    ((11, 'F087'), {'psftype': 'stpsf', 'variable': True, 'nlambda': 1}, (100, 100)),
+    ((12, 'F129'), {'psftype': 'epsf', 'wcs': FakeWCS(), 'nlambda': 1}, None),
+    ((13, 'F087'), {'psftype': 'epsf', 'variable': True, 'nlambda': 1}, (100, 100)),
```
schlafly: Is the chromatic = True case in here somewhere?  It seems to have been removed?

scripts/romanisim-make-image
schlafly:
```suggestion
'An optional FITS file to read to get an array of counts to add into the simulated image.  '
'Useful for wrapping idealized images.  '
'If 2 arguments are sent in, then the second argument is assumed to be the HDU to use (default=0).'
```

scripts/romanisim-make-l3
```diff
@@ -66,7 +70,10 @@ if __name__ == '__main__':
 
     metadata = deepcopy(parameters.default_mosaic_parameters_dictionary)
     if args.date is not None:
-        metadata['basic']['time_mean_mjd'] = Time(args.date, format='isot').mjd
+        metadata['coadd_info']['time_mean'] = Time(args.date, format='isot')
```
schlafly: Ugh, good catch!

romanisim/psf.py
```diff
@@ -13,16 +13,137 @@
...
+        'ROMAN.META.EXPOSURE.START_TIME': date.isot
+    }
+    ref_paths = getreferences(header, reftypes=['epsf'], observatory='roman')
+    model = datamodels.open(ref_paths['epsf'])
+
+    return model
+
+
+@cache
+def get_gridded_psf_model(psf_ref_model, oversample=None):
```
schlafly:
```suggestion
def get_gridded_psf_model(psf_ref_model, oversample=None, focus=0, spectral_type=1):
```

romanisim/psf.py
```diff
@@ -13,16 +13,137 @@
...
+    reference files in CRDS.
+    The input reference files have 3 focus positions and this is using
+    the in-focus images. There are also three spectral types that are
+    available and this code uses the M5V spectal type.
+    """
+    # Open the reference file data model
+    # select the infocus images (0) and we have a selection of spectral types
+    # A0V, G2V, and M6V, pick G2V (1)
+    focus = 0
+    spectral_type = 1
```
schlafly:
```suggestion
```

schlafly (commented):
Thanks Jonathan.  A handful of minor things, and one major thing I don't understand in the meshgrid treatment.  See the comments in line---thanks!

.github/workflows/tests.yml
```diff
@@ -25,12 +37,22 @@ jobs:
     with:
       envs: |
         - linux: check-style
+  latest_crds_contexts:
+    uses: ./.github/workflows/contexts.yml
```
zacharyburnett:
```suggestion
  latest_crds_contexts:
    uses: spacetelescope/crds/.github/workflows/contexts.yml@main
    with:
      observatory: roman
```

zacharyburnett (commented):
with the new change to the CRDS repository (https://github.com/spacetelescope/crds/pull/1127) we can directly call their `contexts.yml` workflow again and only query `roman`, instead of vendorizing it here (the issue before was that HST CRDS would occasionally go down, which would block everything)

.github/workflows/tests_extra.yml
zacharyburnett:
```suggestion
      cache-restore-keys: |
        ${{ needs.latest_data_cache.outputs.cache_key }}
      cache-key: ${{ needs.latest_data_cache.outputs.cache_key }}-${{ needs.crds_context.outputs.context }}
```

zacharyburnett (commented):
this change _should_ cache CRDS by making a new cache with it included; we can test that beforehand at https://github.com/stscieisenhamer/romanisim/pull/1, or just merge that and test it here in this PR

this change adds the CRDS context to the end of `cache-key` ensuring that the FIRST cache restore attempt will fail because it won't exist; HOWEVER, `cache-restore-keys` is a fallback for that situation, so it will restore without CRDS and then store the files in `/tmp/data/crds/`, and then during the job cleanup, the cache will be saved with the new name (with the CRDS context appended)

.github/workflows/tests.yml
```diff
@@ -41,17 +65,25 @@ jobs:
...
         - linux: py311-xdist
         - linux: py312-xdist
+        - linux: py313-xdist
         # `tox` does not currently respect `requires-python` versions when creating testing environments;
         # if this breaks, add an upper pin to `requires-python` and revert this py3 to the latest working version
-        - linux: py3-cov-xdist
+        - linux: py313-cov-xdist
           coverage: codecov
-        - macos: py3-xdist
+        - macos: py313-xdist
```
zacharyburnett: remove the `-xdist`s to attempt to solve CRDS sync issues
```suggestion
       - linux: py311-oldestdeps-cov
       - linux: py311
       - linux: py312
        - linux: py313
       # `tox` does not currently respect `requires-python` versions when creating testing environments;
       # if this breaks, add an upper pin to `requires-python` and revert this py3 to the latest working version
        - linux: py313-cov
         coverage: codecov
        - macos: py313
```

romanisim/psf.py
```diff
@@ -171,49 +390,123 @@ def make_psf(sca, filter_name, wcs=None, stpsf=True, pix=None,
...
+    x_0, y_0 : float or None
+        Position to calculate the psf. If None, (0., 0.) is used
+
+    Returns
+    -------
+    psf : nd.array
+        The psf profile.
+    """
+    x_0 = 0. if x_0 is None else x_0
+    y_0 = 0. if y_0 is None else y_0
```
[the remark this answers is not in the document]
stscieisenhamer: Right, since we're not centering on the bounding box thing anymore. Fixing...

schlafly (approved):

PR #260, opened 2025-09-29
Updates to saturation handling, custom config files, read_pattern defaults, WCS handling

schlafly:
This PR contains several miscellaneous changes:
* It pulls in the fix to the default MA pattern handling to allow usage of non-default MA tables from @npadmana 's PR.
* It adds a new "none" reference type.  This is different from None in that it will not trigger pulling from CRDS when --usecrds is set; normally reference files set to None are populated from CRDS when --usecrds is set.  This enables disabling particular CRDS reference files in a user configuration file.
* It adds an optional saturation argument to the nonlinearity class.  If set, the counts are clipped to the nonlinearity level before the nonlinearity correction is applied.  The goal here is to avoid doing non-linearity corrections outside the allowed range of the linearity reference file.  
* It computes the "inverse saturation" level using the saturation level and the linearity object, and passes this to the inverse linearity object when it's constructed, so that likewise we can avoid computing inverse linearity relationships outside their allowed range.
* It adds the PA argument to the galsim.roman WCS call; this was previously only supported in CRDS mode.
* It updates some tests to work correctly when not using CRDS; the new PA code relies on the pa_aperture metadata being set, and this was missed in some test code.

PR #262, opened 2025-10-07
Improve saturation handling

schlafly:
#260 introduced an issue where saturated pixels were being mishandled.  This PR changes the handling of saturated pixels in two ways:
1. It does not require the inverse linearity and linearity functions to invert in order for the linearity and inverse linearity simulations to proceed.  I wanted to avoid doing math where the polynomials will go crazy, but if we adjust the saturation limit to achieve that romancal will have to know about it in its processing, or these pixels will fail.  So we might as well continue failing on these pixels and push the problem to fixing the reference files.
2. I was capping the inverse & linearity corrections at the saturation level so that we did not try to reach values past saturation.  When the inverse saturation & saturation limits didn't quite match, this could lead us to cap slightly before saturation and never reach it, leading to weird problematic ramps.  Now we cap at 10% larger than the estimated saturation / inverse saturation limits, which seems to address most of the issue.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -706,8 +706,7 @@ def gather_reference_data(image_mod, usecrds=False):
             if np.any(m):
                 log.warning(
                     f'{np.sum(m)} points with problematic saturation / inverse linearity '
-                    'values; setting saturation of these points to 10 DN!')
-            inv_saturation[m] = 10 * u.DN
+                    'values!')
         else:
             inv_saturation = None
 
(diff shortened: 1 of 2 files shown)


romanisim/image.py
```diff
@@ -717,7 +716,12 @@ def gather_reference_data(image_mod, usecrds=False):
             ilin_model.coeffs[:, nborder:-nborder, nborder:-nborder].copy(),
             ilin_model.dq[nborder:-nborder, nborder:-nborder].copy(),
             gain=out['gain'],
-            saturation=inv_saturation)
+            saturation=inv_saturation * 1.1)
+        # fudge factor of 10% on the inverse saturation ensures that we'll
+        # continue to fill up pixels above their nominal saturation limits
+        # so that we can robustly see these pixels as saturated.
+        # It relies on the linearity polynomials not going immediately
```
PaulHuwe, line 723: Do we have any indication of how likely this is?

PaulHuwe (approved):
LGTM with one question

PR #269, opened 2025-10-21
Add default visit_file_group/sequence/activity

schlafly:
The move from the maker utilities to the create_fake_data routines led to some different defaults for the ggsaa metadata.  The default value of -99999 doesn't fit into the smallint / tinyint, leading to database issues.  This changes romanisim to force those to small values.


PaulHuwe (approved):
LGTM

PR #283, opened 2025-12-05
reorder cache key factors to prevent duplicate cache creation

zacharyburnett:
should resolve the issue where adding the CRDS cache to the existing data cache would create a new cache every time

Since the `restore-keys` functionality is prefix-based, this _prepends_ the `crds-<context>` factor BEFORE the desired key to make sure it isn't restored and then appended to by accident


schlafly (approved):
Thank you!

PR #284, opened 2025-12-05
Use backward transform rather than numerical inverse again

### LINKED ISSUE #202: Fix for `test_wcs` failure
This call: https://github.com/spacetelescope/romanisim/blob/a874a4a077fa072110468cc9a12f981cf7b1cebc/romanisim/wcs.py#L307
was broken (`.numerical_inverse` API) by gwcs 0.24 in the way that it is being used here, but `invert` works perfectly fine tor the test in question.

`.invert` will correctly call the `numerical_inverse` if an analytic inverse cannot be found. Typically, the analytic inverse is preferred. @schlafly indicated that at some point the analytic inverse was incorrect; hence, this usage. If this has been fixed I suggest switching to `.invert` here.

A PR restoring the GWCS functionality is forthcoming, so a `.invert` change is not "required" at this time.

schlafly:
The distortion reference file had an issue where it didn't invert correctly.  This has been fixed, and the inverse distortion polynomial is much faster to evaluate than computing its inverse numerically.  So this switches romanisim to use the backward transform again.

Closes #202.

### DIFF
--- a/romanisim/ris_make_utils.py
+++ b/romanisim/ris_make_utils.py
@@ -94,7 +94,7 @@ def set_metadata(meta=None, date=None, bandpass='F087', sca=7,
     if usecrds:
         try:
             context = api.get_default_context('roman')
-        except:
+        except crds.ServiceError:
             context = None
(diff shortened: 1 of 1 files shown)


schlafly:
Tests are failing due to Gaia time-outs; merging.

PR #290, opened 2025-12-16
Convert Quantity inputs to the units of the coordinate frame

nden:
When. the inputs to a GWCS object are quantities, GWCS converts to the units of the corresponding frame. There was a bug in gwcs where this unit conversion was not done and this is fixed now in https://github.com/spacetelescope/gwcs/pull/660

### DIFF
--- a/romanisim/tests/test_wcs.py
+++ b/romanisim/tests/test_wcs.py
@@ -57,6 +57,11 @@ def test_wcs():
                        [cc2.ra / galsim.degrees, cc2.dec / galsim.degrees])
     pos2 = wcsgalsim.toImage(cc1)
     assert np.allclose([pos.x, pos.y], [pos2.x, pos2.y])
+
+    # Using direct unit input to Galsim we should get identical results
+    pos3 = wcsgalsim.toImage(cc1.ra.deg, cc1.dec.deg, units="deg")
+    assert np.all((pos2.x, pos2.y) == pos3)
+ 
     # also try some arrays
     xx = np.random.uniform(0, 4096, 100)
     yy = np.random.uniform(0, 4096, 100)
--- a/romanisim/wcs.py
+++ b/romanisim/wcs.py
@@ -293,10 +293,12 @@ class GWCS(galsim.wcs.CelestialWCS):
             return r, d
 
     def _xy(self, ra, dec, color=None):
-        # _xy accepts ra/dec in radians; we decorate r1, d1 appropriately.
-        r1 = np.atleast_1d(ra) * u.rad
-        d1 = np.atleast_1d(dec) * u.rad
+        # _xy accepts ra/dec in radians.
+        # Convert ra/dec to degrees before passing them to the WCS
+        # which defines the output units to be "deg".
 
+        r1 = np.rad2deg(np.atleast_1d(ra))
+        d1 = np.rad2deg(np.atleast_1d(dec))
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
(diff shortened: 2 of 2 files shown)


WilliamJamieson (approved):

schlafly:
This looks good.  I should wait for the gwcs change to go in before merging?

WilliamJamieson:
I was waiting to coordinate with you on this. If you're good with these changes all start merging the gwcs stuff.

schlafly:
Yes, this looks fine.  Maybe just so I understand the bigger picture: gwcs plans to no longer support calling on coordinate objects (the toImage change), and the removal of units is moving away from astropy units, which can be problematic?  

I would have claimed that the radians -> degree conversion that was being accomplished by adding the radians units before calling the wcs.invert routine must have been happening, since I hadn't seen huge errors in placing things in images?

WilliamJamieson:
> Yes, this looks fine. Maybe just so I understand the bigger picture: gwcs plans to no longer support calling on coordinate objects (the toImage change), and the removal of units is moving away from astropy units, which can be problematic?
> 
You misunderstand the issue. GWCS is working towards a consistent API not changing support for calling on coordinate objects or anything like that.

What is happening is now if you pass an object with units into GWCS you get an object with units out of GWCS. Prior to this change when the way the code in question operated was that it was effectively returning the output of the GWCS's backward transform model with only modifications due to footprints not units. By the construction of the models involved with the gwcs you are creating in romanisim in this case that backward transform does not attach units to its output even when its input has units. This is due to a limitations inherent in astropy. The galsim code does not work with units directly, so you end up with a unit conversion error which is basically "cannot turn a dimensioned thing (has a unit) into something dimensionless (no unit)".

GWCS does in fact "know" the units of what the inverse should be (unlike the direct astropy models); namely, the units of the input frame. Thus with the API consistency it looks at the outputs of the backward transform and converts them to or attaches the units of the input frame. Then when galsim does its bit it runs into an issue with the units. An alternative solution is to strip the units from the gwcs.inverse's output prior to passing it galsim.

> I would have claimed that the radians -> degree conversion that was being accomplished by adding the radians units before calling the wcs.invert routine must have been happening, since I hadn't seen huge errors in placing things in images?

romanisim/tests/test_wcs.py
```diff
@@ -55,8 +55,11 @@ def test_wcs():
     cc2 = wcsgalsim2.toWorld(pos)
     assert np.allclose([cc1.ra / galsim.degrees, cc1.dec / galsim.degrees],
                        [cc2.ra / galsim.degrees, cc2.dec / galsim.degrees])
-    pos2 = wcsgalsim.toImage(cc1)
-    assert np.allclose([pos.x, pos.y], [pos2.x, pos2.y])
+    # The units parameter is the desired units of  the output
+    # The GWCS object's output_frame has units of "deg" so it expects
+    # the input to be in degrees.
+    pos2 = wcsgalsim.toImage(cc1.ra.deg, cc1.dec.deg, units="deg")
+    assert np.allclose([pos.x, pos.y], pos2)
```
WilliamJamieson:
```suggestion
    pos2 = wcsgalsim.toImage(cc1)
    assert np.allclose([pos.x, pos.y], [pos2.x, pos2.y])

    # Using direct unit input to Galsim we should get identical results
    pos3 = wcsgalsim.toImage(cc1.ra.deg, cc1.dec.deg, units="deg")
    assert np.all((pos2.x, pos2.y) == pos3)
 
```

WilliamJamieson (changes requested):

WilliamJamieson (approved):

PR #292, opened 2025-12-21
Remove astropy units for electron and DN

schlafly:
This PR removes astropy units usage for electron, DN, and related units (electron/s, DN/s, electron/DN gain) from romanisim codebase. 

romanisim originally did not use these units, but we at some point added them to the Roman reference files and romancal.  We later removed them.  Since they are no longer in the reference files it now makes sense to remove them again in romanisim.

This PR is mostly unit removals and the removal of related code that tries to correctly handle cases when units are or are not sent into a function.  I've tried to add more information and docstrings tracking intended units to partially make up for the removed programmatic unit requirements.  I've made corresponding updates to the tests.

### DIFF
--- a/romanisim/image.py
+++ b/romanisim/image.py
@@ -61,28 +61,30 @@ def make_l2(resultants, read_pattern, read_noise=None, gain=None, flat=None,
     Parameters
     ----------
     resultants : np.ndarray[nresultants, ny, nx]
-        resultants array
+        Resultants array in DN
     read_pattern : list[list] (int)
-        list of lists of indices of reads entering each resultant
-    read_noise : np.ndarray[ny, nx] (float)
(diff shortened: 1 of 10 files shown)


braingram:
Does this need related romancal changes? I'm seeing test failures in the stcal downstream testing:
https://github.com/spacetelescope/stcal/actions/runs/20969999648/job/60270581555#step:10:709
which I think is due to romancal expecting a value to be a quantity:
https://github.com/spacetelescope/romancal/blob/6299939de14c70bb61766c7c038deb02b047ee0f/romancal/source_catalog/tests/test_injection.py#L202

schlafly:
Yes, I feel like I missed that we were depending on romanisim main and just merged this.  We only need to delete the .value.  Sorry, will address tomorrow.

PR #315, opened 2026-01-29
Improve docstring clarity for `read_catalog`

cgarling:
I was confused what the purpose of the `coord` argument was in the case that the user provides a simple catalog file with source positions already defined. This clarifies the function of some of the arguments in `read_catalog` for different input `filename`.


schlafly (approved):
Thank you!

PR #327, opened 2026-02-16
RCAL-1299: Support for Regression Testing

PaulHuwe:
The bandpasses were being stored in an unsorted fashion, which lead to random position in a random number generator. This caused inconsistent results from run to run with the same seed. This ticket rectifies that.

### DIFF
--- a/romanisim/catalog.py
+++ b/romanisim/catalog.py
@@ -352,6 +352,10 @@ def make_cosmos_galaxies(coord,
     source_pert = np.ones(len(sim_ids))
     source_pert += ((0.2) * rng_numpy.normal(size=len(sim_ids)))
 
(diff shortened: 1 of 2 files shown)


schlafly (approved):
This looks fine.  For consideration, if instead we just deleted the set(...) wrapping the definition of BANDPASSES, would that resolve this issue?  I think I infer that the set(...) is the thing that's giving us different orderings on different runs.

PaulHuwe:
Except that my testing is sending bandpasses, not using the default - though I agree that changing the default BANDPASSES is a good idea.

PaulHuwe:
I did further testing, and changing the type of BANDPASSES wasn't enough. (Though I did that change.)

schlafly (approved):

PR #328, opened 2026-02-18
romanisim.models

yuedong0607:
This PR integrates the `galsim.roman` functionality into `romanisim` and enables direct access to both CRDS and the Roman Technical Information repository:
- Introducing a new module `romanisim.models`
- Replacing corresponding `galsim.roman` calls throughout: `romanisim `codebase, executables,  and unit tests.

Current status:

- [x] - passed all existing unit tests
- [x] - passed two end-to-end tests for making l1 and l2 images: 
```
   romanisim-make-image --level 1 --usecrds --psftype epsf --rng_seed 42 out_l1.asdf
   romanisim-make-image --level 2 --usecrds --psftype epsf --rng_seed 42 out_l2.asdf
```
   - The outputs are consistent with previous behavior, differing only at the level of background modeling.
   - Full agreement is achieved when `models.bandpass` and `models.background` calls are reverted to the original `galsim.roman` implementations.

TODO (near term):
- **Decide on a strategy for handling the hardcoded path to the locally cloned `roman-technical-information` repository.**
- Add unit tests for `romanisim.models`
- Add `romanisim.models.flat`
- Improve docstrings and example notebook


yuedong0607:
> But it seems the build tests still complain.
It seems it has captured only the RuntimeError exception which leads to the complaints. I replaced them with Exception instead.

romanisim/models/parameters.py
```diff
@@ -0,0 +1,343 @@
...
+    "distortion": None,
+    "flat": None,
+    "gain": 2,  # electron/DN
+    "inverselinearity": None,
+    "linearity": None,
+    "integralnonlinearity": None,
+    "readnoise": 5.0,  # DN
+    "saturation": 55000,  # DN
+    "ipc": None,
+}
```
schlafly: How much of this do you think we should keep vs. how much duplicates stuff you have elsewhere now?

romanisim/models/read_noise.py
```diff
@@ -0,0 +1,165 @@
...
+        metadata : dict, optional
+            Metadata overrides to apply before CRDS lookup. If provided, values
+            are merged into the model metadata tree.
+
+        """
+
+        if image_mod is not None:
+            ref_file = crds.getreferences(
+                image_mod.get_crds_parameters(),
+                reftypes=["dark", "gain"],
```
schlafly: Probably needs to be read_noise.  I'd suggest making something like lines 85-106 a new function in utils that gets a specific set of reference files for an image_mod?  I feel like there's a lot of duplicated code in the different modules along these lines.

schlafly (commented):
Hi Yuedong,

This was a ton of work!  Thanks.

I left a few comments in line that we should discuss.  Broadly:
- we should think about how the current parameter defaults in parameters.reference_data interact with the new defaults
- it probably makes sense to move some of the logic for grabbing a reference file using either default information or the image model to a separate routine?  It looks very similar among the different models.
- is there any documentation on the new data files added in this PR?  I'm okay with adding them but I know that I personally can't maintain them; I don't really know where they come from, etc..  Some mixture of documentation & agreement that you're on the hook to maintain those at some level would be appreciated.  On the flip side, we're on the hook to make sure we can do releases when you need them for your sims?
- Special treatment of persistence relative to other instrumental effects is because that one needs more than just a reference file?

yuedong0607:
> Hi Yuedong,
> 
> This was a ton of work! Thanks.
> 
> I left a few comments in line that we should discuss. Broadly:
> 
> * we should think about how the current parameter defaults in parameters.reference_data interact with the new defaults
> * it probably makes sense to move some of the logic for grabbing a reference file using either default information or the image model to a separate routine?  It looks very similar among the different models.
> * is there any documentation on the new data files added in this PR?  I'm okay with adding them but I know that I personally can't maintain them; I don't really know where they come from, etc..  Some mixture of documentation & agreement that you're on the hook to maintain those at some level would be appreciated.  On the flip side, we're on the hook to make sure we can do releases when you need them for your sims?
> * Special treatment of persistence relative to other instrumental effects is because that one needs more than just a reference file?

Hi Eddie,

Thanks for the careful review and helpful comments! I'll work through your inline comments shortly. Regarding your general points:
* Agreed. For now, I've just tried to retain all existing default parameters from the three sources: (following the priority order) 1) roman_techinical_information repository; 2) romanisim; 3) galsim.roman
* I agree there is some repetitive across models when gathering reference files. The main complication is that a particular model (e.g, the nonlinearity model) may require more than one reference files along with some pre-processing. So it might be a little difficult for having an universal _get_crds_model() method. But I think it should be easy to have a shared helper routine to centralize the logic for gathering the required CRDS references. Any suggestion on this?
* The newly added data files are all from galsim.roman. I'll ask around see if there's corresponding documentations.
* In roman_imsim, there're some ongoing efforts regarding the up-the-ramp sampling that follow a slightly different logical (e.g., caching a big photon array) compared to the current romanisim implementation. Thus I have not yet attempted to create a unified model for persistence. Additionally, the modeling parameters in roman_technical_information repository appear to differ from those used in both romanisim and galsim.roman. I think maybe we need some further discussion on the details?

romanisim/models/read_noise.py
yuedong0607: It is indeed a typo. Fixed. Thanks for catching it!

romanisim/models/saturation.py
```diff
@@ -0,0 +1,169 @@
...
+        #     # [from roman_imsim] this maybe should be better applied at
+        #     # read time? it's not actually clear to me what the right
+        #     # thing to do is in detail.
+        #     if not isinstance(img, u.Quantity):
+        #         img *= u.DN
+        #     img = np.clip(img, 0 * u.DN, self.saturation_map, out=img)
+
+        #     # m = resultants >= saturation
+        #     # dq[m] |= parameters.dqbits['saturated']
+        #     # return resultants, dq
```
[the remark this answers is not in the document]
yuedong0607: I think that makes sense. If needed, we could implement a separate method in these modules, something like apply_dq().

yuedong0607:
Hi @schlafly , do you think there are any additional changes I should make to this branch before merging to main?

Also, would it be okay for me to start removing `parameters`, `bandpass`, `nonlinearty`, and `wcs` from the `romanisim` folder? I'll double-check the corresponding version in `romanisim.models` contain everything needed before doing so.

PR #340, opened 2026-03-10
Add images in new filters to regtests

schlafly:
This PR adds a couple new uncal files to the regtest data made for romancal in order to better support testing the new multiband catalog as part of https://github.com/spacetelescope/romancal/pull/2051 .


braingram (commented):
Commands LGTM.

Is there a romancal PR updating https://github.com/spacetelescope/romancal/blob/main/romancal/scripts/make_regtestdata.sh to use these? Once that's ready would you regenerate the regtest files with the 2 new scripts and run the romancal regtests?

braingram (approved):
Thanks for running and linking the tests!

PR #341, opened 2026-03-10
Skip too-large FFTs.

schlafly:
galsim 2.8 changed what happens when galaxies need to be rendered that require too-large FFTs.  Before it triggered an error which we caught and turned into a warning, skipping the galaxy processing.  Now it triggers a warning but proceeds to render the requested, very expensive galaxy. This sets a galsim configuration parameter to restore the old behavior.

See https://github.com/GalSim-developers/GalSim/blob/v2.8.4/CHANGELOG.rst#api-changes for details on the galsim change.

### DIFF
--- a/romanisim/__init__.py
+++ b/romanisim/__init__.py
@@ -13,6 +13,9 @@ rate images that are analagous to typical astronomical images.
 
 from importlib.metadata import version
 import logging
+import galsim
+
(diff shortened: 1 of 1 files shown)


braingram (approved):
Looks reasonable to me.

If I'm following, this should work with older galsims (so no need to bump that version) since with those it will just set some unused parameter (and the old behavior is what we want anyway).

PR #343, opened 2026-03-23
Support specifying nexposures.

schlafly:
This PR adds functionality to romanisim-make-image and romanisim-make-l3 to allow metadata keywords to be specified on the command line.  It then uses that functionality in romanisim-make-stack to determine the correct number of exposures entering each visit for an APT-generated program.

The mechanism used is that romanisim-make-image and romanisim-make-l3 gain a new --meta argument that takes arguments like visit.nexposures=4.  The visit.nexposures=4 string is parsed to interpret this as a metadata keyword and a value for that keyword.  This will fail if we have any weird metadata that has, for example, '=' in the name, but I think it will handle all existing cases in a reasonable way.

romanisim-make-stack computes the number of exposures in each visit by looping through the CSV file and explicitly counting each combination of (plan, pass, segment, observation, visit), and then adds --meta visit.nexposures=N to each  line of the output commands file.

### DIFF
--- a/romanisim/ris_make_utils.py
+++ b/romanisim/ris_make_utils.py
@@ -1,6 +1,7 @@
 """Miscellaneous utility routines for the simulation file maker scripts.
 """
 
+import ast
 from copy import deepcopy
 import os
 import pathlib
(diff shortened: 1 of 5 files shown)


romanisim/ris_make_utils.py
```diff
@@ -49,6 +50,48 @@ def merge_nested_dicts(dict1, dict2):
...
+        Metadata dict to update in-place.
+    """
+    if args.meta is None:
+        return
+    for item in args.meta:
+        key, _, value = item.partition('=')
+        try:
+            value = ast.literal_eval(value)
+        except (ValueError, SyntaxError):
+            pass
```
PaulHuwe: Do you want to throw a warning about the `skipped` meta entry?

PaulHuwe (approved):
Thanks - checked out. LGTM.

PR #346, opened 2026-04-09
Support more WCS types

schlafly:
Folks doing simulations outside romanisim may want to use WCSes other than the romanisim GWCS, gwcs, or galsim.roman.WCS.  This adds support for a generic FITS header.


PaulHuwe (approved):
LGTM

PR #350, opened 2026-04-16
Fix readnoise pixfrac scaling

### LINKED ISSUE #349: `--effreadnoise` CLI option is silently ignored
**Package:** romanisim 0.13.1
**File:** `romanisim/l3.py`, function `simulate()`, lines ~417-418
**Severity:** The `--effreadnoise` parameter on `romanisim-make-l3` has no effect. When passed, it silently sets the effective read noise to zero instead of using the provided value.

## Description

The `romanisim-make-l3` CLI accepts `--effreadnoise` to override the default (pessimistic) read noise estimate. The value is passed through to `l3.simulate()`. However, in `simulate()`, the logic is:

```python
# l3.py, simulate(), around line 109
if effreadnoise is None:
    # ... compute pessimistic CDS-based estimate ...
    effreadnoise = effreadnoise * etomjysr
else:
    effreadnoise = 0    # <-- BUG: user's value is discarded
```

When `effreadnoise is not None` (i.e., the user passed a value), the code sets it to **0** instead of using the provided value. This zero is then passed to `simulate_cps()`, which adds no read noise to the image and sets `var_rnoise = 0`.

Note that `simulate_cps()` itself handles `effreadnoise` correctly — if it receives a non-None, non-zero value, it adds Gaussian noise scaled by that value and sets `var_rnoise = effreadnoise**2`. The bug is only in `simulate()` which wraps `simulate_cps()`.

## Expected behavior

The `else` branch should pass the user's value through:

```python
else:
    pass  # use the value as provided (already in MJy/sr)
```

or equivalently, just delete lines 128-129.

## Impact

There are two issues:

1. **The `--effreadnoise` override doesn't work.** The user-provided value is silently discarded and replaced with zero. The flag exists to allow users to supply a realistic read noise estimate, but it has no effect.

2. **The default (None) estimate is pessimistic.** When `--effreadnoise` is not passed, `simulate()` computes a CDS-based estimate assuming only 2 reads. This overestimates the read noise by ~9× in amplitude (~80× in variance) compared to a real exposure with ~40 reads and ramp fitting. The code comments acknowledge this and direct the user to `--effreadnoise`, which doesn't work due to issue #1.

In our testing, the pessimistic default caused romancal SourceCatalogStep to detect only 6/130 sources on the idealized L3. With the read noise corrected to the realistic pipeline value (via monkey-patch), detection improved to 50/130 — matching the L2 direct detection result.

## Reproducer

```bash
# Generate with default (pessimistic) noise — this works
romanisim-make-l3 --bandpass F158 --radec 10 0 --npix 1000 --exptime 100 \
  --rng_seed 42 test_default.asdf catalog.ecsv

# Generate with explicit effreadnoise — value is silently ignored, noise = 0
romanisim-make-l3 --bandpass F158 --radec 10 0 --npix 1000 --exptime 100 \
  --effreadnoise 0.035 --rng_seed 42 test_override.asdf catalog.ecsv

# Check:
python -c "
import roman_datamodels as rdm
import numpy as np
m1 = rdm.open('test_default.asdf')
m2 = rdm.open('test_override.asdf')
print(f'Default var_rnoise:  {np.nanmedian(m1.var_rnoise):.4e}')  # ~0.67
print(f'Override var_rnoise: {np.nanmedian(m2.var_rnoise):.4e}')  # 0.0 (should be ~0.0012)
"
```

## Workarounds

**1. Use the bug to get a zero-noise image, then add noise yourself.** Since passing any `--effreadnoise` value results in zero read noise, you can generate a noiseless L3 and add correctly-scaled Gaussian noise externally.

**2. Call `l3.simulate_cps()` directly** — it accepts `effreadnoise` correctly and applies it as intended.

**3. Monkey-patch `simulate_cps`** before calling `l3.simulate()`:

```python
_original = l3.simulate_cps
target_effreadnoise = 0.035  # MJy/sr

def patched(*args, effreadnoise=None, **kwargs):
    if effreadnoise is not None and effreadnoise == 0:
        effreadnoise = target_effreadnoise
    return _original(*args, effreadnoise=effreadnoise, **kwargs)

l3.simulate_cps = patched
result = l3.simulate(..., effreadnoise=target_effreadnoise, ...)
l3.simulate_cps = _original
```

schlafly:
Nikhil and Greg's investigations of the romanisim L3 files indicated that they were not as deep as they should be.  The root cause is that the read noise scaling with the pixel scale fraction was missing a factor of the square of the pixel scale fraction.

The default read noise computation uses the read noise per read and imagines that the total read noise is just sqrt(2) times the per-read read noise, a pessimistic assumption.  It then converts that to electrons / s using the total exposure time times the square of the pixscalefrac, reasoning that the total exposure time needs to be split up among pixscalefrac ** 2 output pixels.  The result is then converted to MJy / sr units for use downstream.

The final conversion to MJy / sr units assumed that we were starting with e / s / coadd pixel units, but the computation had been done assuming e / s / native pixel units, introducing a spurious factor of the pixscalefrac squared.  This PR fixes that issue.

It also fixes an unrelated error where the effective read noise was zeroed if an explicit value was sent by the user, makes some mild documentation updates, and adds a test of the read noise scaling with pixel scale fraction.

Closes https://github.com/spacetelescope/romanisim/issues/349 .

### DIFF
--- a/romanisim/l3.py
+++ b/romanisim/l3.py
@@ -398,9 +398,10 @@ def simulate(shape, wcs, efftimes, filter_name, catalog, nexposures=1,
     # Set effective read noise
     if effreadnoise is None:
         readnoise = np.median(romanisim.parameters.reference_data['readnoise'])
+        # read noise in DN
         gain = np.median(romanisim.parameters.reference_data['gain'])
         effreadnoise = (
-            np.sqrt(2) * readnoise * gain)
+            np.sqrt(2) * readnoise * gain)  # electron, difference of two reads
         # sqrt(2) from subtracting one read from another
         effreadnoise /= (np.median(efftimes * pixscalefrac ** 2) / nexposures)
         # divided by the typical exposure length
@@ -412,10 +413,10 @@ def simulate(shape, wcs, efftimes, filter_name, catalog, nexposures=1,
         # note that we are ignoring all of the individual reads, which also
         # contribute to reducing the effective read noise.  Pass --effreadnoise
         # if you want to do better than this!
-        effreadnoise = effreadnoise * etomjysr  # electron -> MJy/sr
+        effreadnoise = effreadnoise * etomjysr * pixscalefrac ** 2  # electron -> MJy/sr
+        # factor of pixfrac ** 2 because etomjysr is per L3 pixel, but
+        # this effreadnoise calculation is per native pixel
         # converting to MJy/sr units
-    else:
-        effreadnoise = 0
 
     chromatic = False
     if (len(catalog) > 0 and not isinstance(catalog, astropy.table.Table)
@@ -481,7 +482,7 @@ def simulate_cps(image, filter_name, efftimes, objlist=None, psf=None,
                  xpos=None, ypos=None, coord=None, sky=0, bandpass=None,
                  effreadnoise=None, maggytoes=None, etomjysr=None,
                  rng=None, seed=None, ignore_distant_sources=10,):
-    """Simulate average MegaJanskies per steradian in a single SCA.
+    """Simulate average MJy/sr in a single SCA.
 
     Parameters
     ----------
@@ -507,10 +508,10 @@ def simulate_cps(image, filter_name, efftimes, objlist=None, psf=None,
     effreadnoise : float
         Effective read noise for mosaic (MJy / sr)
     maggytoes: float
-        Factor to convert electrons to MJy / sr; one maggy makes
+        Factor to convert e/s to MJy / sr; one maggy makes
         this many e/s.
     etomjysr : float
-        Factor to convert electron to MJy/sr;  one e/s/pix corresponds
+        Factor to convert e/s to MJy / sr; one e/s/coadd pix corresponds
         to this MJy/sr.
     rng : galsim.BaseDeviate
         random number generator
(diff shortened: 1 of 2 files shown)


schlafly:
@npadmana , mind taking a look at this and seeing if it addresses the issues you found?

npadmana:
I took a very quick look at this. 

This significantly improves the detections, although the pessimistic/default case seems to be worse than what I get with a single L2 mapping directly to an L3.  My understanding is that difference comes from the up the ramp reads, while the default version assumes 2 reads. Is that right?

schlafly:
I think that's part of it.  It's certainly true that the assumption we're making for the default effective read noise is the maximally pessimistic one.

The larger issue is probably with how the resampling is simulated.  For the single image case, pixfrac = 0.5 case, romancal invents 4 pixels for each single input pixel.  The errors accordingly get correlated.

What I'm trying to simulate (modulo more bugs) in romanisim is that in your nominal N seconds you have taken 4 exposures each of N/4 in length and interleaved them.  So your uncertainties are uncorrelated but you pay more read noise.  I think that for a single exposure you wouldn't really want to use pixscalefrac = 0.5.  I'll need to think more about what I would expect the noise to look like for the single exposure romancal case.

PR #355, opened 2026-04-20
do not build for free-threaded Python

### LINKED ISSUE #354: photutils 3.0.0 incompatible with free-threading
Builds are hitting a failure when using photutils 3 and thread-free mode:
https://github.com/spacetelescope/romanisim/actions/runs/24589627314/job/72124049432?pr=328#step:9:6670

@zacharyburnett , I think the easiest thing to do would be to remove the thread-free mode?  I think I like that more than pinning to photutils < 3?  I think you added the thread-free build; do you have a sense here?

zacharyburnett:
fixes #354

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,11 +14,11 @@ jobs:
         FFTW_DIR: /opt/homebrew/opt/fftw/lib/
       targets: |
         # Linux wheels
-        - cp3*-manylinux_x86_64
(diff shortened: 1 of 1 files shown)


braingram (approved):
Nice find! The docs failure looks like hopefully just a network hiccup for https://lsstdesc.org/Coord/_build/html/objects.inv (which resolves for me).

PR #363, opened 2026-06-23
[SCSB-237] build Stable ABI3 wheels with `Py_LIMITED_API`

zacharyburnett:
Resolves [SCSB-237](https://jira.stsci.edu/browse/SCSB-237)

building against the Limited API means we can build just ONE wheel (in this case for Python 3.11) and it should work on all future Python versions: https://docs.python.org/3/c-api/stable.html#limited-c-api

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,26 +1,71 @@
 name: build
 
 on:
+  pull_request:
   release:
     types: [released]
-  pull_request:
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
 
 jobs:
   build:
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@2835f0cacddf3f8de198db9afdb5354a5cebe0ef # v2.6.3
-    with:
-      env: |
-        FFTW_DIR: /opt/homebrew/opt/fftw/lib/
-      targets: |
-        # Linux wheels
-        - cp3*[!t]-manylinux_x86_64
-        # MacOS wheels
-        - cp3*[!t]-macosx_x86_64
-        # MacOS arm64 wheels
-        - cp3*[!t]-macosx_arm64
-      sdist: true
-      test_command: python -c "from romanisim import ramp_fit_casertano"
-      upload_to_pypi: ${{ (github.event_name == 'release') && (github.event.action == 'released') }}
-    secrets:
-      pypi_token: ${{ secrets.PYPI_PASSWORD_STSCI_MAINTAINER }}
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-latest
+          - macos-latest
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-tags: true
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3"
+      - if: ${{ runner.os == 'Linux' }}
+        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
+        with:
+          platforms: all
+      - uses: pypa/cibuildwheel@294735312765b09d24a2fbec22660ce817587d55 # v4.1.0
+        with:
+          package-dir: ./
+          output-dir: dist/
+      - run: pip install build
+      - run: python -m build --sdist
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: dist-${{ runner.os }}-${{ runner.arch }}
+          path: ./dist/
+  publish:
+    if: (github.event_name == 'release') && (github.event.action == 'released')
+    needs: [build]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      attestations: write
+    # To add required reviewers before deployment,
+    # edit the protection rules in GitHub Settings:
+    # Settings > Environments > pypi > Add required reviewers
+    environment:
+      name: pypi
+      url: https://pypi.org/p/romanisim
+    steps:
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: dist*
+          path: dist/
+          merge-multiple: true
+      - uses: actions/attest@a1948c3f048ba23858d222213b7c278aabede763 # v4.1.1
+        with:
+          subject-path: "dist/*"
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -1,26 +1,71 @@
...
+  cancel-in-progress: true
+
+permissions: {}
+
 jobs:
   build:
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@2835f0cacddf3f8de198db9afdb5354a5cebe0ef # v2.6.3
-    with:
-      env: |
-        FFTW_DIR: /opt/homebrew/opt/fftw/lib/
```
zacharyburnett, line 14: @spacetelescope/romanisim-maintainers this build seems to work WITHOUT defining `FFTW_DIR: /opt/homebrew/opt/fftw/lib/` like before; the test command (`python -c 'from romanisim import ramp_fit_casertano'`) completes successfully, but is there another test command we can apply to the wheel to make sure it's compiled with `FFTW`?

pyproject.toml
```diff
@@ -113,5 +113,10 @@ exclude_lines = [
 [tool.ruff.lint]
 extend-ignore = ["E501"]
 
+[tool.cibuildwheel]
+build = "cp311-*"  # build for CPython 3.11 Stable API
+skip = ["cp314t-*"]  # explicitly skip free-threaded builds
+test-command = "python -c 'from romanisim import ramp_fit_casertano'"
```
zacharyburnett: we define the `cibuildwheel` configuration here, which makes the build workflow more generalized (and updatable via the package template)

.github/workflows/build.yml
```diff
@@ -1,26 +1,71 @@
...
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3"
+      - if: ${{ runner.os == 'Linux' }}
+        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
+        with:
+          platforms: all
+      - uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084 # v3.4.1
```
braingram: Why 3.4.1 and not one of the newer versions?

.github/workflows/build.yml
```diff
@@ -1,26 +1,71 @@
...
+      - uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+        with:
+          pattern: dist*
+          path: dist/
+          merge-multiple: true
+      - uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: "dist/*"
+      # To upload to PyPI without a token, add this workflow file as a Trusted Publisher in the project settings on the PyPI website
+      - uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
```
braingram: This is out-of-date too. Would you update the action versions?

braingram (commented):
abi3audit on the mac wheel gives a 👎 

```
[16:34:14] 👎 romanisim-0.14.1.dev6+g097407ae0-cp311-cp311-macosx_11_0_arm64.whl: ramp_fit_casertano.abi3.so uses the Python 3.11 ABI, but is tagged for the Python 3.2 ABI
           ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
           ┃ Symbol                           ┃ Version ┃
           ┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
           │ PyImport_ImportModuleLevelObject │ 3.7     │
           │ PyImport_GetModule               │ 3.8     │
           │ PyErr_SetExcInfo                 │ 3.7     │
           │ PyBuffer_Release                 │ 3.11    │
           │ PyErr_GetExcInfo                 │ 3.7     │
           │ PyType_FromModuleAndSpec         │ 3.10    │
           │ PyInterpreterState_Get           │ 3.9     │
           │ PyType_GetQualName               │ 3.11    │
           │ PyErr_SetHandledException        │ 3.11    │
           │ PyMemoryView_FromMemory          │ 3.7     │
           │ PyUnicode_FindChar               │ 3.7     │
           │ PyExc_ModuleNotFoundError        │ 3.6     │
           │ PyInterpreterState_GetID         │ 3.7     │
           │ PyIndex_Check                    │ 3.8     │
           │ PyCMethod_New                    │ 3.9     │
           │ PyModule_NewObject               │ 3.7     │
           │ PyObject_GetBuffer               │ 3.11    │
           │ PyObject_GenericSetDict          │ 3.7     │
           │ PyObject_GenericGetDict          │ 3.10    │
           │ Py_Version                       │ 3.11    │
           │ PyType_GetSlot                   │ 3.4     │
           │ PyUnicode_Substring              │ 3.7     │
           │ PyModuleDef_Init                 │ 3.5     │
           │ PyUnicode_GetLength              │ 3.7     │
           └──────────────────────────────────┴─────────┘
           💁 romanisim-0.14.1.dev6+g097407ae0-cp311-cp311-macosx_11_0_arm64.whl: 1 extensions scanned; 1 ABI version mismatches and 0 ABI violations found
```

Is that being run in the CI? If not we should add that before claiming abi3 compatibility. I think if we use a later cibuildwheel it will automatically run abi3audit (starting in 4.0.0):
https://github.com/pypa/cibuildwheel/releases/tag/v4.0.0

.github/workflows/build.yml
schlafly: The full test suite definitely exercises this, but you want something minimal?  Something like:
import galsim
galsim.fft.fft2(np.zeros((10, 10))
should be enough to trigger it, though I haven't tried rebuilding without FFTW!

zacharyburnett:
running `abi3audit` on the wheels locally now prints out nothing; is that expected?
image

pyproject.toml
```diff
@@ -113,5 +113,16 @@ exclude_lines = [
...
+
 [tool.cibuildwheel.macos]
-before-build = "brew install eigen fftw"
+before-all = "brew install eigen fftw"
+
+[tool.cibuildwheel.macos.environment]
+FFTW_DIR = "/opt/homebrew/opt/fftw/lib/"
+
+[tool.cibuildwheel.linux]
+before-all = "yum install -y fftw-devel || apk add fftw"  # manylinux is AlmaLinux; musllinux is Alpine
```
zacharyburnett, line 128: we can define the FFTW installation here and also pass the required environment variable for macOS homebrew

zacharyburnett:
`cibuildwheel@4.1.0` is now auditing the ABI3 wheel, and testing with `galsim.fft.fft2()` ensures the presence of FFTW in the build environment

braingram (approved):
Thanks! This LGTM. With the new cibuildwheel abi3audit is now running in the CI.

I pulled down the built wheel and tested it locally (with `pytest --pyargs romanisim`) and didn't see any failures that indicated wheel problems. I did see failures related to crds access (since I don't have a local setup yet) and some about galsim missing input data (missing COSMOS_25.2_training_sample, etc) since I also don't have that locally.

zacharyburnett:
well at least ONE of these C extension packages Just Worked ™️ with ABI3 😓 the rest of them will require some heavy lifting if we want more ABI3 wheels

schlafly:
Thanks both!

PR #368, opened 2026-07-08
remove the subtree of roman-technical-information

yuedong0607:
The roman-techinical-information is now pip installable in version 1.5: [link](https://github.com/RomanSpaceTelescope/roman-technical-information/releases/tag/v1.5)
So we can remove the local copy from data folder.


romanisim/models/parameters.py
```diff
@@ -82,32 +79,32 @@
...
+#         data_path = os.path.join(data_path, "data", "roman-technical-information")
+#         return Path(data_path)
+#     except Exception as e:
+#         raise RuntimeError(
+#             "Could not locate reference data. "
+#             f"Set environment variable {ENV_VAR} to the roman-technical-information directory."
+#         ) from e
 
-roman_tech_repo_path = get_reference_data_root()
+# roman_tech_repo_path = get_reference_data_root()
```
schlafly: Let's just delete this, I think?  Do you see a reason for us to keep it?  If I understand correctly, with your changes we no longer support it.

schlafly (commented):
Awesome, glad that the design let this be removed easily.  One minor comment inline.  Sorry for being slow getting to this!

romanisim/models/parameters.py
yuedong0607: Hi Eddie, sorry I just noticed this comment. I've removed this helper function.

yuedong0607:
The failed style check looks wired to me. It is reporting issues in the files that I didn't touch. Could it be related to the ruff version now 0.16.0?

schlafly:
Yes.  Don't worry about that.  I'll fix in a separate PR and merge, thanks.

PR #370, opened 2026-07-16
Use differential velocity aberration by default

schlafly:
Prior to this PR, differential velocity aberration was only computed when scale_factor was given a negative value.  By default, a trivial value of 1 was taken.  romanisim-set-image already overrode this value, so this makes no change for romanisim-make-image invocations.  romanisim-make-stack invocations however were calling set_metadata without a particular value, getting the 1 default, and not using differential velocity aberration.  This fixes that case and makes users who just want to get the WCS (for example) get dva effects by default.


PaulHuwe (approved):
LGTM

PR #371, opened 2026-07-23
Add reference read

schlafly:
This PR adds the ability for romanisim to make files with the reference read subtracted and data_encoding_offset added to the science resultants.  It also updates the documentation and tests accordingly.

Reference reads are created during apportion_reads_to_resultants to get the correct pedestal level and inverse linearity application.  Reference reads are not given CRs or pesistence because Roman reference reads are usually reset reads and it seems to me that the effective exposure times for these reads will be quite different than for normal reads.


schlafly:
Here's an image showing the difference between simulation runs with and without this change:
image
I'm showing data1 - (data2 + reference_read - data_encoding_offset).  The expectation is that this difference is mostly zero.  It's literally zero on the border pixels which are not simulated and don't have noise.  It's got minor read noise differences elsewhere since the newly simulated reference read has read noise.  That checks out.

PR #377, opened 2026-08-14
Fix header sent to CRDS for epsf

### LINKED ISSUE #376: CRDS ePSF integration
`romanisim.psf.get_epsf_from_crds()` builds its CRDS header with `"ROMAN.META.INSTRUMENT.DETECTOR": f"SCA{sca:02d}"` (e.g. `"SCA01"`). The `roman_wfi_epsf` CRDS rmap's `Match` selector, and the `roman.meta.instrument.detector` field actually stored inside every epsf reference file, use `"WFI01"`--`"WFI18"` instead. `"SCA01"` never matches any SCA-specific entry, so CRDS's `Match` selector silently falls through to the generic `('N/A', <filter>)` wildcard entry and incorrectly returns the same PSF reference file regardless of which SCA was requested. E.g.,

```python
import crds
from astropy.time import Time

date = Time("2026-01-01T00:00:00")

for detector in ("SCA01", "WFI01"):
    header = {
        "ROMAN.META.INSTRUMENT.NAME": "wfi",
        "ROMAN.META.INSTRUMENT.DETECTOR": detector,
        "ROMAN.META.INSTRUMENT.OPTICAL_ELEMENT": "F087",
        "ROMAN.META.EXPOSURE.START_TIME": date.isot,
    }
    print(detector, "->", crds.getrecommendations(header, reftypes=["epsf"], observatory="roman"))
```

produces 

```
SCA01 -> {'epsf': 'roman_wfi_epsf_0120.asdf'}
WFI01 -> {'epsf': 'roman_wfi_epsf_0172.asdf'}
```

I'm filing this as a romanisim issue because the problematic code lives here, but there's a question of whether "SCA01" should resolve to the same thing as "WFI01" **inside the CRDS call**, so this could alternatively get fixed in CRDS rather than romanisim.

The problematic code is in `romanisim/psf.py`, function `get_epsf_from_crds` (line 344 specifically):

```python
header = {
    "ROMAN.META.INSTRUMENT.NAME": "wfi",
    "ROMAN.META.INSTRUMENT.DETECTOR": f"SCA{sca:02d}",
    "ROMAN.META.INSTRUMENT.OPTICAL_ELEMENT": galsim2roman_bandpass[filter_name],
    "ROMAN.META.EXPOSURE.START_TIME": date.isot,
}
ref_paths = getreferences(header, reftypes=["epsf"], observatory="roman")
```

Versions used:
- romanisim commit `15baa6581702f83d3a88b4766fea446163d83d55` (also reproduces on PyPI 0.14.0 wheel)
- crds 13.2.7 (also on the current `spacetelescope/crds` `main` branch)
- CRDS context resolved: `roman_0058.pmap` (server default at time of testing)

cgarling:
Closes #376

### DIFF
--- a/romanisim/psf.py
+++ b/romanisim/psf.py
@@ -341,7 +341,7 @@ def get_epsf_from_crds(sca, filter_name, date=None):
         )
     header = {
         "ROMAN.META.INSTRUMENT.NAME": "wfi",
-        "ROMAN.META.INSTRUMENT.DETECTOR": f"SCA{sca:02d}",
+        "ROMAN.META.INSTRUMENT.DETECTOR": f"WFI{sca:02d}",
         "ROMAN.META.INSTRUMENT.OPTICAL_ELEMENT": galsim2roman_bandpass[
             filter_name
         ],
--- a/romanisim/tests/test_psf.py
+++ b/romanisim/tests/test_psf.py
@@ -5,6 +5,7 @@ import pytest
 
 import numpy as np
 from romanisim import psf
+from romanisim.models.bandpass import galsim2roman_bandpass
 import galsim
 import galsim.roman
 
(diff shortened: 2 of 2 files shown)


schlafly (approved):
Yuck!  Good catch, thank you!

schlafly:
This is fixed, but adding a little more context...
https://github.com/spacetelescope/romanisim/pull/377/changes#diff-0116eef8f0db8dad591a7e68e2b1788c620e044614f0a741337349487f0f28c1R344
is the correct fix.

Each WFI detector has two different numbers, one that starts with SCA and then is ~five digits and corresponds to some kind of manufacturing ID, and one that starts with WFI and is two digits and refers to the location of the detector in the focal plane.  Because the detectors are SCAs and the overall instrument is called WFI, it's easy to ~incorrectly say something like SCA01 (mixing the SCA id and the WFI id nomenclature).  The current policy is to never refer to SCAXY and to instead always refer to WFIXY.  stpsf previously referred to SCAXY when it meant WFIXY, though now it accepts either SCAXY or WFIXY as synonyms always meaning WFIXY (preserving backward compatibility with its previous behavior.

The previous behavior asking for an non-existent SCAXY reference file was incorrect and this is the right solution.

PR #382, opened 2026-09-01
Use `psf_noipc` array from CRDS epsf reference

### LINKED ISSUE #381: IPC convolution in CRDS ePSF
On line 371 of psf.py the `psf` array of a CRDS epsf reference file is read and used to construct the GriddedPSFModel. https://github.com/spacetelescope/romanisim/blob/fbf12144bd1ae7930909d96a7ba01b983949e329/romanisim/psf.py#L357-L371
This path is taken when `psftype="epsf"` and `usecrds=True`. 

Looking at the current CRDS epsf reference files (I specifically looked at `roman_wfi_epsf_0182.asdf` for WFI01 and F158 filter), it contains both `psf` and `psf_noipc` arrays. Per the file's description ("roman/meta/description" entry):

> The psf and extended_psf arrays have been convolved with an interpixel capacitance (IPC) kernel to make them more like empirical PSFs (ePSFs) that will be observed on-orbit. The IPC kernel comes  ground testing at Goddard Space Flight Center and are the same kernels available in CRDS. For use by Roman I-Sim, a version of both the PSF postage stamps and the extended PSF image are 
provided with no IPC kernel applied. These IPC-free versions of the PSF postage stamps and extended PSF are stored in the psf_noipc and extended_psf_noipc arrays, respectively.

It therefore seems to me by using the `psf` array, romanisim may be double-applying the IPC, because it is reading the IPC-convolved array from the reference file and then also applying its own IPC in the L1 and L2 image simulation paths. Specifically I think this occurs for `image.simulate` calls for L1 and L2. I think the fix is just reading `psf_noipc` instead of `psf`.

cgarling:
Fixes #381

### DIFF
--- a/docs/romanisim/refs.rst
+++ b/docs/romanisim/refs.rst
@@ -44,6 +44,11 @@ EPSF
 The EPSF reference provides pre-determined PSF models for all detectors and nine
 different locations on each detector. For rendering at any given position, the
 PSF is interpolated between the nearest calculated PSFs from the reference file.
+The reference file stores the PSF stamps both with and without an interpixel
+capacitance (IPC) kernel applied (the ``psf`` and ``psf_noipc`` arrays).
+romanisim applies IPC to the resultants during L1 simulation (see
+:doc:`l1`), so it renders scenes with the IPC-free ``psf_noipc`` array to
+avoid applying IPC twice.
 
 Flat field
 ----------
--- a/romanisim/psf.py
+++ b/romanisim/psf.py
@@ -364,11 +364,16 @@ def get_gridded_psf_model(
     The input reference files have 3 focus positions and this is using
     the in-focus images. There are also three spectral types that are
     available and this code uses the M5V spectal type.
+
+    The ``psf`` array in the reference file has already been convolved with an
+    interpixel capacitance (IPC) kernel; the ``psf_noipc`` array has not.
+    romanisim applies IPC to the resultants in ``romanisim.l1.make_l1``, so we
+    use the IPC-free ``psf_noipc`` array here to avoid applying IPC twice.
     """
     # Open the reference file data model
     # select the infocus images (0) and we have a selection of spectral types
     # A0V, G2V, and M6V, pick G2V (1)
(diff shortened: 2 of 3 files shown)


schlafly:
@tddesjardins , this PR tells me that the PSF reference file is wrong in another way.  Here Chris is taking the 'noipc' extension and convolving it with the IPC kernel directly to get the IPC-applied PSF.  But the IPC kernel links _native_ pixels together, rather than oversampled pixels together, so the convolution should be something more like 
```
Kd = np.zeros((2 * ov + 1, 2 * ov + 1))
Kd[::ov, ::ov] = K
psf = ndimage.convolve(psf_noipc, Kd)
```
This probably gets subsumed into broader discussions about making the PSF reference file correct, but what we have in there now has a broken IPC application.

schlafly:
Thanks Chris.  This looks good.  As I mentioned here
https://github.com/spacetelescope/romanisim/pull/382#issuecomment-5512424039
though I think the old test was broken and what you are really showing is that the CRDS PSF reference file isn't handling IPC right either (so romanisim wasn't applying it right, and it wasn't applied right in the file, etc.---this is an embarrassment of errors).  I've updated the test to do what I think is the right convolution.  That test now fails and I've marked it 'xfail.'  There are some discussions to improve the PSF reference file and this will have to be part of them.  We can discuss what photometry tests 'should' be doing in light of the misbehavior here separately.

tddesjardins:
> @tddesjardins , this PR tells me that the PSF reference file is wrong in another way. Here Chris is taking the 'noipc' extension and convolving it with the IPC kernel directly to get the IPC-applied PSF. But the IPC kernel links _native_ pixels together, rather than oversampled pixels together, so the convolution should be something more like
> 
> ```
> Kd = np.zeros((2 * ov + 1, 2 * ov + 1))
> Kd[::ov, ::ov] = K
> psf = ndimage.convolve(psf_noipc, Kd)
> ```
> 
> This probably gets subsumed into broader discussions about making the PSF reference file correct, but what we have in there now has a broken IPC application.

Yeah, Tim had already flagged this in RTB as an issue to be fixed when we next update the files.

schlafly (approved):

PR #384, opened 2026-09-02
Fix application of IPC

### LINKED ISSUE #383: IPC not applied in `make_l1`
`add_ipc` does not mutate the input resultants, so the return value is the convolved result
https://github.com/spacetelescope/romanisim/blob/fbf12144bd1ae7930909d96a7ba01b983949e329/romanisim/l1.py#L674-L696

However, when this function is called in `make_l1` the output is not captured, so the convolved resultants are discarded and not used.

https://github.com/spacetelescope/romanisim/blob/fbf12144bd1ae7930909d96a7ba01b983949e329/romanisim/l1.py#L783-L786

Similarly, in the `apply` method of the IPC class, the image is not updated in-place as the docstring says, the local variable is rebound to the convolved output of `ndimage.convolve`, but this value is never returned from the `apply` method and so the IPC convolved result is similarly discarded.

https://github.com/spacetelescope/romanisim/blob/fbf12144bd1ae7930909d96a7ba01b983949e329/romanisim/models/ipc.py#L100-L139

Here is a code snippet which should illustrate the problem:

```python
    import numpy as np
    from romanisim.models import parameters, ipc
    from romanisim import l1
    import galsim

    branch = 'add_ipc'
    parameters.n_pix = 21
    y0 = x0 = 10
    read_pattern = [[1], [2], [3]]
    source = 1e6

    # An asymmetric, unit-sum kernel: a transpose or axis flip in the
    # application would change the recovered footprint.
    kernel = np.array([[0.004, 0.020, 0.006],
                       [0.050, 0.860, 0.030],
                       [0.003, 0.024, 0.003]])

    if branch == 'add_ipc':
        # make_l1 falls back to add_ipc with the default kernel
        ipc_model = None
        expected = np.asarray(ipc.ipc_kernel)
    else:
        ipc_model = ipc.IPC(usecrds=False)
        ipc_model.ipc_kernel = kernel
        expected = kernel

    counts = np.zeros((parameters.n_pix, parameters.n_pix), dtype='f4')
    counts[y0, x0] = source
    res, _ = l1.make_l1(
        galsim.Image(counts), read_pattern, read_noise=0,
        pedestal_extra_noise=0, gain=1, crparam=None, saturation=1e30,
        ipc_model=ipc_model, seed=1)

    footprint = res[-1].astype(np.float64)[y0 - 1:y0 + 2, x0 - 1:x0 + 2]
    footprint -= parameters.pedestal

    # This should be 9, as flux from the central pixel should be redistributed 
    # by the IPC kernel, but it is 1 on main, and so fails, because the neighbors
    # have not received any flux from the IPC convolution
    assert np.count_nonzero(footprint) == 9
```

I am working on a PR but fixing this by updating in-place seems to break one of the cosmic ray tests because activating the IPC convolution redistributes single-pixel CR flux into the neighboring pixels

cgarling:
Fixes #383. 

In-place mutation is indicated in docstring of `IPC.apply`, but we need to return the array as well for use in `make_l1` as `add_ipc(resultants)` does not mutate in-place, so we need both branches to return their values. Also added `mode` and `cval` options to `add_ipc` to support both 1) the case in `make_l1` which convolves with resultants which carry the reset pedestal, and 2) the case of convolving a PSF-stamp (or other flux-only array) with the IPC which should use zero-padded convolution. 

# Outstanding issues

Fixing the application of IPC creates issues for two existing tests: `test_make_l1_and_asdf` performs a check that the resultant sits at the pedestal everywhere except on cosmic rays, but with IPC now applied, the CR flux leaks into neighbor pixels which are *not* jump-flagged, so the existing test failed. I fixed this by effectively disabling IPC by using an identity convolution kernel. Basically the same issue is present in `test_image_input`, but that uses the `simulate()` entry point which doesnt have an `ipc_model` keyword argument, so there I just set `crparam=None` to disable it. You may want to do something more principled here.

### DIFF
--- a/docs/romanisim/l1.rst
+++ b/docs/romanisim/l1.rst
@@ -30,6 +30,8 @@ Interpixel capacitance (IPC) is added following non-linearity and before read-ou
 
 This is slightly different than including IPC in the PSF kernel because including IPC in the PSF kernel leaves the Poisson noise uncorrelated between pixels.
 
+The resultants do not include the reference pixels, so the convolution treats the pixels just outside the array as having the same values as the pixels on its edge.  The IPC also spreads the charge from a cosmic ray into its neighbors, so the jump flags are grown to cover the IPC kernel.
+
 Persistence
 -----------
 Persistence is implemented in the simulator following Sanchez+2023.  This follows
--- a/romanisim/l1.py
+++ b/romanisim/l1.py
@@ -671,28 +671,32 @@ def read_pattern_to_tij(read_pattern, reference_read=False):
     return tij
 
 
-def add_ipc(resultants, ipc_kernel=None):
+def add_ipc(resultants, ipc_kernel=None, mode='constant', cval=0):
     """Add IPC to resultants.
 
     Parameters
     ----------
     resultants : np.ndarray[n_resultant, ny, nx]
         resultants describing a scene
+    ipc_kernel : np.ndarray[3, 3] or None
+        IPC convolution kernel.  Defaults to ``romanisim.models.ipc.ipc_kernel``.
+    mode, cval
+        Boundary handling, passed to ``scipy.ndimage.convolve``.  The default
+        (``'constant'``, 0) is appropriate when the array carries only flux
+        (e.g. a PSF stamp); callers whose array includes the reset pedestal
+        should pass ``mode='nearest'`` so the border is not pulled toward zero.
 
     Returns
     -------
     np.ndarray[n_resultant, ny, nx]
         resultants with IPC
     """
-    # add in IPC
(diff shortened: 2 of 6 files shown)


schlafly:
I made some mostly documentation changes and changed the various IPC functions to only modify in place and not also return (I see claiming to do both as the thing that led to the bug).  I kept your change to the default handling of the boundary pixels but moved it to the defaults rather than changing the arguments in make_l1.  I expanded the jump flagging by 3x3 to handle the IPC -> CRs are bigger than they should be issue.  Do you want to take a look and then I'll merge?

cgarling:
Yep looks good, thanks!

PR #388, opened 2026-09-10
Handle pixel-convolved and pixel-unconvolved ePSFs

schlafly:
The next set of CRDS ePSF reference files will include IPC and convolution by the native pixel, without IPC-free and pixel-free versions of the PSFs.  romanisim currently depends on the IPC and pixel-free versions.  They will also be in 'native pixel' units rather than 'idealized pixel' units that come from STPSF; that is, they will include distortion.

This PR makes changes to romanisim in order to handle these new reference files and adds associated tests and documentation.

The proof that this all works is that romanisim generated images processed through romancal using the same PSF reference file produce good agreement in centroids.  The level of agreement with this PR is shown below:
image
Agreement is ~0.005 pix per coordinate, good but not extraordinary.  The two leading causes are some approximations in galsim (pixel phase error, reduces with lower maxk_threshold) and the fact that romanisim only linearly approximates PSF variation across the four corners, while romancal uses the full 3x3 reference file grid (spatial dependence across the detector).  But ~0.005 pix / star is pretty good so I think we can be happy with this.

The approach for handling pixel convolution is to add a new pixel_convolved attribute to the PSF object so that add_objects_to_image knows whether to render a source with 'no_pixel' mode or not.  The 'no_pixel' mode makes photon shooting impossible but we don't exercise that mode much anyway.  The mode also makes the L3 PSF simpler since it doesn't need to think about the difference between the output pixel scale (always convolved in) and the native pixel scale (only convolved in by galsim if it's the output pixel scale), which needn't be the same for L3 images.

The approach for handling IPC is to deconvolve the IPC out of the new-style reference files using the IPC kernel, so that the rest of romanisim always sees an IPC-free PSF.  IPC is then added back in during L1 creation.

The approach for handling distortion is to attach a different WCS to the PSF object depending on the type of reference file.  Old style reference files continue to include only a rotation relative to north reflecting the detector alignment and the pixel scale.  New style reference files include the local Jacobian of the WCS where the PSF was evaluated.  

We handle both new and old-style ePSF reference files by keying off the overall normalization of the CRDS reference file images; new-style reference files have normalizations roughly of oversample^2, while old-style have normalizations of roughly 1.

I developed this PR in collaboration with AI / Claude.

### DIFF
--- a/romanisim/psf.py
+++ b/romanisim/psf.py
@@ -404,9 +404,8 @@ def deconvolve_ipc(psf_images, ipc_kernel, oversample, pad=32):
 
         A(f) = sum_ij a_ij exp(-2 pi i oversample (i f_y + j f_x))
 
-    which we divide out.  We build it analytically in order to keep
-    the result independent of whether the
-    stamp has an even or an odd number of samples on a side.
+    which we divide out, being careful to leave the PSF centering
+    unaffected.
 
     romanisim deconvolves using this function and reconvolves later in
     ``romanisim.l1.make_l1`` with the same kernel.  Away from
--- a/romanisim/psf.py
+++ b/romanisim/psf.py
@@ -338,8 +338,7 @@ def get_epsf_from_crds(sca, filter_name, date=None):
 
     override = reference_data.get('epsf')
     if isinstance(override, str):
-        log.info('Using epsf reference %s instead of the CRDS default',
-                 override)
+        log.info('Forcing the use of ePSF reference %s.', override)
         return datamodels.open(override)
 
     if date is None:
@@ -377,6 +376,10 @@ def epsf_is_pixel_convolved(psf_ref_model, focus=0, spectral_type=1):
     older, un-convolved reference is normalized to the enclosed-flux
     fraction and sums to slightly less than one.  We cut at a sum of 1.1.
 
+    A reference file following either convention should land close to one
+    of those two values; we warn if it does not, since that suggests the
+    file follows some third convention that we are guessing about.
+
     Parameters
     ----------
     psf_ref_model : roman_datamodels.EpsfRefModel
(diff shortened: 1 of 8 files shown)


schlafly:
@t-brandt , this is a riff on your ePSF work.  Would you mind reviewing and I'll get this merged?

romanisim/psf.py
```diff
@@ -353,9 +362,100 @@ def get_epsf_from_crds(sca, filter_name, date=None):
...
+
+    Returns
+    -------
+    bool
+        True if the reference PSF has been convolved with the pixel
+        response function.
+    """
+    total = np.sum(psf_ref_model.psf[focus, spectral_type, 0, :, :])
+    return bool(total > 1.1)  # mildly larger than 1 to provide some buffer
+
```
t-brandt: In this case, we expect something close to one or something close to `oversample**2`.  It would be a bit drastic to throw an error if we didn't get something close to one of these, but a warning seems appropriate.  Maybe throw a warning if the sum is not between, say, 0.9 and 1.05 or between `oversample**2*0.9` and `oversample**2**1.05`?

romanisim/psf.py
```diff
@@ -333,6 +336,12 @@ def get_epsf_from_crds(sca, filter_name, date=None):
     """
     from crds import getreferences
 
+    override = reference_data.get('epsf')
+    if isinstance(override, str):
+        log.info('Using epsf reference %s instead of the CRDS default',
+                 override)
```
t-brandt: I think the default will be the ePSF as soon as we can derive it from in-flight data.  Reword info statement to something like "forcing the use of ePSF reference %s?"

t-brandt (commented):
Looks good to me.  My comments are minor.

